Issue #91, opened 2026-04-15
Synthetic pattern-based integration test for generation correctness (single-chunk + rolling)


mmcdermott:
## Summary

Replace the current "does the CLI produce non-empty parquet?" smoke-test style with a real **generation-correctness** integration test: train a tiny model on CPU against a synthetic pattern dataset where we know the ground truth, then assert that generation — both single-chunk and rolling — reproduces the patterns the model was trained on. The point is to validate not just that the plumbing wires up but that the model + generation path actually **works** end-to-end on something we can verify.

**Not for this PR.** Design-only for now; implementation as a follow-up in dev after #86 lands.

## Motivation

Our current tests fall into two camps:

- **Unit / doctest level**: `Model.generate`, `_rolling_generate`, `_truncate_at_first_eos` are exercised on random-init models with deterministic seeds. The doctests prove the **code path** executes and the **shapes** are right. They don't prove that, if the model actually learned something, generation would recover it.
- **CLI integration level**: `test_generate_trajectories_runs` / `test_generate_trajectories_rolling_runs` drive the full `MEICAR_generate_trajectories` CLI and check that parquet files exist, splits line up, and — as of #86's review follow-up — that the rolling path produces output longer than the single-chunk path could. Better than pure existence checks, but still silent on whether the generated content is *right*.

The missing middle: a test that (1) trains a model quickly enough to fit in CI, (2) on a dataset simple enough that correct generation is unambiguous, and (3) exercises the **full generation path** so we catch regressions that break the generation math without breaking any shape or plumbing invariant.

### The pattern dataset

Define a small probabilistic grammar over a tiny vocabulary. Each "program" is one of a few recognizable fixed substrings, emitted with stochastic interleaving:

```
program_A := "A B C D"     # weight 0.4
program_B := "1 2 3"       # weight 0.3
program_C := "R S T U V W X Y Z"  # weight 0.3

sequence := [program] ("|" [program])* EOS
```

The separator `"|"` between programs lets the model learn "after |, pick any program" while still being able to complete a program once it's started. Vocab size: ~20 tokens (A-D, 1-3, R-Z, `|`, `BOS`, `EOS`, plus a pad). Each program is short enough that completing it is a strong test of one-step lookahead.

**Why this shape of task:**
- **Correct generation is unambiguously verifiable.** After seeing `A B`, the model must output `C` — there's no ambiguity, no distributional metric needed. Grammar adherence is a binary per-step check.
[...]

PR #86, opened 2026-04-14
feat: rolling sliding-window generation (EOS + max_new_tokens)

mmcdermott:
## Summary

First slice of #82. Adds a sliding-window generation path to \`Model.generate\` that can emit more tokens than the model's positional-embedding limit would allow in a single HF \`generate()\` call. When \`max_new_tokens\` is left unset the legacy single-chunk behavior is preserved exactly; when it is set, a new \`_rolling_generate\` loop takes over and repeatedly feeds the right-aligned tail of \`(input ++ generated)\` (truncated to \`rolling_context_size\`) back into the model until either every sample in the batch has emitted the EOS token (\`TIMELINE//END\`) or the total new-token budget is exhausted.

**Scope:** EOS + step-cap stopping only. The **time-budget** stopping criterion from the design in #82 is intentionally left to a follow-up so this PR stays small and reviewable.

## Implementation

- \`Model.generate\` gains two new kwargs: \`max_new_tokens: int | None = None\` and \`rolling_context_size: int | None = None\`. When \`max_new_tokens\` is \`None\`, behavior is identical to today. When it is set, the call is dispatched to \`Model._rolling_generate\`.
- \`Model._rolling_generate\` maintains a running \`generated\` buffer and a per-sample \`finished\` mask. Each iteration takes the right-aligned tail of \`(batch.code ++ generated)\` truncated to \`rolling_context_size\` (defaulting to \`max_seq_len - 1\`), calls \`HF_model.generate\` for \`min(remaining_budget, max_seq_len - prompt_len)\` new tokens, zeros out any tokens produced for already-finished rows so they cannot "resurrect", appends the new tokens to the buffer, and updates \`finished\` from any row that contains \`eos_id\`.
- \`Model._truncate_at_first_eos\` (new static helper) takes the final buffer and replaces every token strictly after each row's first EOS with pad. This uses a shifted cumsum so the first EOS itself is preserved.
- \`MEICARModule.predict_step\` now forwards \`self.generation_kwargs\` to \`model.generate\`. The \`generate_trajectories\` CLI populates that dict from a new \`rolling_generation\` config block before \`trainer.predict\` runs.
- \`configs/_generate_trajectories.yaml\` grows a \`rolling_generation.{max_new_tokens, rolling_context_size}\` block, both null by default. Existing generation runs are therefore behavior-preserving.

## Tests (same style as the current suite)

- **Doctests in \`src/MEDS_EIC_AR/model/model.py\`:**
[...]

PR #93, opened 2026-04-15
test: synthetic-grammar integration test for generation correctness

mmcdermott:
## Summary

Adds a single new integration test file, `tests/test_pattern_generation.py`, that trains a tiny `Model` on CPU against a hand-built pattern grammar and verifies that both single-chunk and rolling generation recover the grammar. Closes the test gap identified in #91.

## What's in the file

- **Grammar.** Three fixed programs — \`A B C D\`, \`1 2 3\`, \`R S T U V W X Y Z\` — emitted stochastically with a \`SEP\` token between them. No terminating EOS in the training distribution (see *Why no EOS* below).
- **GrammarFSM.** Deterministic state machine that walks a token sequence and flags any grammar-invalid transition. Has its own doctests covering the three states (\`BETWEEN\`, \`(IN, prog, pos)\`, and the invalid sink).
- **Model fixture.** Module-scoped fixture that constructs a small \`Model\` (2 layers, hidden 32, intermediate 64, \`max_seq_len=16\`, vocab 20) and trains it for 400 Adam steps at lr=3e-3 with a batch size of 32. Training uses \`Model.forward\` directly via \`Mock\` batches — no MEDS fixtures.
- **Two integration tests** against the trained model:
  1. \`test_trained_model_single_chunk_generation_recovers_grammar\` — for each program-start token (and a few mid-program prefixes), asserts that greedy single-chunk generation exactly completes the in-progress program and then emits \`SEP\`.
  2. \`test_trained_model_rolling_generation_preserves_grammar\` — drives \`Model.generate\` with \`max_new_tokens=50\` on a \`max_seq_len=16\` model, spies on \`HF_model.generate\` to assert the rolling loop iterated at least 10 times, and walks every generated token through the \`GrammarFSM\` to assert every transition is grammar-valid.

## Why no EOS in the training distribution

Early iterations of this test trained the model on sequences that ended with EOS. That worked for the single-chunk test but silently broke the rolling test: the trained model learned to emit EOS around token 14, before the rolling loop had a chance to cross a sliding-window boundary. The rolling test would terminate early with a 14-token output, and the real integration signal — "rolling generation stays grammar-valid across multiple sliding windows" — was smothered.

The fix is to train on sequences that are packed full of programs separated by SEPs, padded but never terminated. The model's \`eos_token_id\` is set to an out-of-grammar dummy token (\`DUMMY_EOS = 18\`) so the rolling-generation validation (which rejects \`None\` and pad-collision) is satisfied, but the model never learns to emit that token greedily. Rolling generation is bounded solely by \`max_new_tokens\`, and the full budget of 50 tokens is emitted — verified empirically to trigger ~40 inner \`HF_model.generate\` calls on the pre-submission test run.

## Scope

This is a **test-only change**. Single new file under \`tests/\`. No changes to \`src/\`, no changes to \`conftest.py\`, no new fixtures outside the test file, no new dependencies. Branches off \`dev\` which already has the rolling-generation changes from #86.

## Test plan

- [x] \`uv run pytest tests/test_pattern_generation.py -v\` passes locally (4 tests: 2 doctests + 2 integration). Runtime: ~41s on a laptop CPU.
- [x] \`uv run pytest -q\` full suite passes locally (51 tests total, ~75s).
- [ ] CI passes on this branch.

### DIFF
--- a/tests/test_pattern_generation.py
+++ b/tests/test_pattern_generation.py
@@ -1,0 +1,383 @@
+"""Integration test: train a tiny ``Model`` on a synthetic pattern grammar and verify that
+generation — both single-chunk and rolling — recovers the grammar.
+
+All other tests in this repo either exercise the generation code path on a random-init model
+(doctests, ``test_generate_trajectories.py``'s direct unit tests) or drive the full CLI but only
+check parquet shape. This file fills the missing middle: a real generation-correctness test that
+(1) trains a tiny ``Model`` on CPU in a handful of seconds, (2) has unambiguous ground truth because
+the training distribution is a finite-state grammar, and (3) asserts grammar adherence via a
+deterministic FSM walk of the generated tokens.
+
+The grammar has three programs — ``A B C D``, ``1 2 3``, ``R S T U V W X Y Z`` — emitted
+stochastically with a ``|`` separator between them. Training sequences are packed full of programs
+until no more fit, then padded — **no terminating EOS token**. That's intentional: if we trained
+the model to emit EOS, the rolling-generation test would terminate early (the model would emit EOS
+before the sliding window had a chance to cross a boundary) and the real integration signal —
+"rolling generation stays grammar-valid across multiple sliding windows" — would be smothered by
+early termination. Instead, we set the model's ``eos_token_id`` to an out-of-grammar dummy token
+and let ``max_new_tokens`` control termination for the rolling test. The single-chunk test
+allows ``SEP`` as the transition token out of a program (which the model will emit, since it's
+what the training distribution always does at that position).
+
+The whole file fits in under a minute of CPU training on a small model and uses no MEDS fixtures;
+it talks to ``Model`` directly via mock batches.
+"""
+
+from __future__ import annotations
+
+import random
+from unittest.mock import Mock
+
+import pytest
+import torch
+
+from MEDS_EIC_AR.model.model import Model
+
+# ---------------------------------------------------------------------------
+# Grammar vocabulary
+# ---------------------------------------------------------------------------
+
+PAD = 0
+SEP = 1
+# Three programs, each a fixed token sequence. Tokens are picked to be mnemonic: A..D as 2..5 (the
+# four-token program, analogous to "A B C D"), 1..3 as 6..8 (the three-token program), and R..Z as
+# 9..17 (the nine-token program).
+PROGRAMS: dict[str, tuple[int, ...]] = {
+    "A": (2, 3, 4, 5),
+    "B": (6, 7, 8),
+    "C": (9, 10, 11, 12, 13, 14, 15, 16, 17),
+}
+PROGRAM_NAMES = tuple(PROGRAMS.keys())
+PROGRAM_WEIGHTS = (0.4, 0.3, 0.3)
+PROGRAM_STARTS = {name: prog[0] for name, prog in PROGRAMS.items()}
+
+# Reserved dummy EOS index outside the training distribution. The model is configured with
+# ``eos_token_id=DUMMY_EOS`` so the rolling-generation validation (which rejects ``None`` and pad
+# collisions) is satisfied, but the model is never trained on sequences containing ``DUMMY_EOS`` so
+# it will never emit this token greedily. Rolling generation is bounded solely by ``max_new_tokens``.
+DUMMY_EOS = 18
+VOCAB_SIZE = 20
+MAX_SEQ_LEN = 16
+
+
+def _start_token_to_name(token: int) -> str | None:
+    for name, start in PROGRAM_STARTS.items():
+        if token == start:
+            return name
+    return None
+
+
+# ---------------------------------------------------------------------------
+# Grammar FSM
+# ---------------------------------------------------------------------------
+
+
+class GrammarFSM:
+    """Deterministic walker over the pattern grammar. Each ``step(token)`` returns either the new state name
+    or ``None`` (meaning the transition is grammar-invalid).
+
+    States:
+      - ``"BETWEEN"`` — between programs (or at the very start). Valid next tokens: any
+        program-start.
+      - ``("IN", prog_name, pos)`` — inside program ``prog_name`` having just emitted position
+        ``pos``. Valid next: either ``prog[pos + 1]`` if there is one, or (at the final position)
+        ``SEP``.
+
+    No terminal state: the training distribution never emits an EOS token (see module docstring),
+    so the grammar is effectively infinite — any valid sequence of programs separated by ``SEP``
+    can continue indefinitely.
+
+    The FSM is used for two things in this file:
+
+    1. Replaying a ground-truth prompt through the FSM to establish the initial state before we
+       call the model (so the test can feed the model into the middle of a program and assert it
+       completes that program correctly).
+    2. Walking the model-generated continuation token-by-token and asserting every transition is
+       valid.
+
+    >>> fsm = GrammarFSM()
+    >>> fsm.step(2)  # start of program A
+    ('IN', 'A', 0)
+    >>> fsm.step(3)
+    ('IN', 'A', 1)
+    >>> fsm.step(4)
+    ('IN', 'A', 2)
+    >>> fsm.step(5)
+    ('IN', 'A', 3)
+    >>> fsm.step(SEP)
+    'BETWEEN'
+    >>> fsm.step(6)  # start of program B
+    ('IN', 'B', 0)
+    >>> fsm.step(8)  # would skip B[1]=7, invalid
+    >>> fsm.state is None
+    True
+    """
+
+    def __init__(self) -> None:
+        self.state: str | tuple[str, str, int] | None = "BETWEEN"
+
+    def step(self, token: int) -> str | tuple[str, str, int] | None:
+        """Transition on ``token``. Returns new state, or ``None`` on invalid transition.
+
+        Once the FSM enters the invalid state (``self.state is None``), further ``step`` calls
+        remain in ``None``.
+        """
+        if self.state is None:
+            return None
+        if self.state == "BETWEEN":
+            name = _start_token_to_name(token)
+            if name is None:
+                self.state = None
+            else:
+                self.state = ("IN", name, 0)
+            return self.state
+        # ("IN", name, pos)
+        _, name, pos = self.state
+        prog = PROGRAMS[name]
+        if pos + 1 < len(prog):
+            if token == prog[pos + 1]:
+                self.state = ("IN", name, pos + 1)
+            else:
+                self.state = None
+            return self.state
+        # At last position of program; next must be SEP.
+        if token == SEP:
+            self.state = "BETWEEN"
+        else:
+            self.state = None
+        return self.state
+
+    def walk(self, tokens) -> int:
+        """Walk a token sequence.
+
+        Returns the index of the first invalid token, or ``len(tokens)``
+        if every transition was valid.
+        """
+        for i, tok in enumerate(tokens):
+            if self.step(int(tok)) is None:
+                return i
+        return len(tokens)
+
+
+# ---------------------------------------------------------------------------
+# Grammar sampling
+# ---------------------------------------------------------------------------
+
+
+def sample_sequence(rng: random.Random, max_len: int = MAX_SEQ_LEN) -> list[int]:
+    """Sample one grammar-valid sequence that fits in ``max_len`` tokens, no terminating EOS.
+
+    Packs programs separated by ``SEP`` until the next full program wouldn't fit. The sequence
+    always ends with a ``SEP`` followed by nothing — i.e. the caller pads with ``PAD`` up to
+    ``max_len``, and ``Model._forward``'s cross-entropy loss ignores those pad positions so the
+    model never sees a "what comes after the final SEP" target. That keeps the trained grammar
+    effectively non-terminating.
+
+    >>> rng = random.Random(0)
+    >>> seq = sample_sequence(rng)
+    >>> GrammarFSM().walk(seq) == len(seq)
+    True
+    >>> seq[-1] == SEP
+    True
+    """
+    tokens: list[int] = []
+    while True:
+        prog_name = rng.choices(PROGRAM_NAMES, weights=PROGRAM_WEIGHTS, k=1)[0]
+        prog = PROGRAMS[prog_name]
+        # Reserve room for the SEP after this program.
+        if len(tokens) + len(prog) + 1 > max_len:
+            break
+        tokens.extend(prog)
+        tokens.append(SEP)
+    if not tokens:  # pragma: no cover — only triggered if max_len is absurdly small
+        tokens.extend(PROGRAMS["B"])
+        tokens.append(SEP)
+    return tokens
+
+
+def sample_batch(rng: random.Random, batch_size: int, max_len: int = MAX_SEQ_LEN) -> torch.Tensor:
+    seqs = [sample_sequence(rng, max_len) for _ in range(batch_size)]
+    pad_to = max(len(s) for s in seqs)
+    padded = torch.full((batch_size, pad_to), PAD, dtype=torch.long)
+    for i, s in enumerate(seqs):
+        padded[i, : len(s)] = torch.tensor(s, dtype=torch.long)
+    return padded
+
+
+def mock_batch(code: torch.Tensor) -> Mock:
+    """Build a minimal ``MEDSTorchBatch``-shaped mock that ``Model`` can consume."""
+    return Mock(code=code, PAD_INDEX=PAD, mode="SM")
+
+
+# ---------------------------------------------------------------------------
+# Trained model fixture
+# ---------------------------------------------------------------------------
+
+
+_NUM_TRAIN_STEPS = 400
+_BATCH_SIZE = 32
+_LEARNING_RATE = 3e-3
+
+
+@pytest.fixture(scope="module")
+def grammar_trained_model() -> Model:
+    """Train a tiny ``Model`` on the pattern grammar.
+
+    Module-scoped so both tests share it.
+    """
+    torch.manual_seed(0)
+    rng = random.Random(0)
+
+    model = Model(
+        {
+            "num_hidden_layers": 2,
+            "num_attention_heads": 2,
+            "hidden_size": 32,
+            "intermediate_size": 64,
+            "max_position_embeddings": MAX_SEQ_LEN,
+            "vocab_size": VOCAB_SIZE,
+            "eos_token_id": DUMMY_EOS,
+        },
+        precision="32-true",
+    )
+    model.train()
+
+    optimizer = torch.optim.Adam(model.parameters(), lr=_LEARNING_RATE)
+
+    for _ in range(_NUM_TRAIN_STEPS):
+        batch_codes = sample_batch(rng, _BATCH_SIZE)
+        batch = mock_batch(batch_codes)
+        optimizer.zero_grad()
+        loss, _ = model(batch)
+        loss.backward()
+        optimizer.step()
+
+    model.eval()
+    return model
+
+
+# ---------------------------------------------------------------------------
+# Generation tests
+# ---------------------------------------------------------------------------
+
+
+# For each program, a short prompt that should unambiguously constrain the next several tokens.
+# The first token is the program's start; we assert greedy decoding completes the program, then
+# transitions via SEP or EOS (both FSM-valid).
+_PROMPTS_AND_EXPECTED_COMPLETIONS: list[tuple[list[int], list[int]]] = [
+    # ``A`` prompt: start of program A. Must continue with A[1], A[2], A[3].
+    ([PROGRAMS["A"][0]], list(PROGRAMS["A"][1:])),
+    # Mid-program: first two tokens of A. Must continue with A[2], A[3].
+    (list(PROGRAMS["A"][:2]), list(PROGRAMS["A"][2:])),
+    # ``B`` prompt.
+    ([PROGRAMS["B"][0]], list(PROGRAMS["B"][1:])),
+    # ``C`` prompt — the long one. Tests that the model has learned the full 9-token sequence.
+    ([PROGRAMS["C"][0]], list(PROGRAMS["C"][1:])),
+    # Mid-C: enough context to pin down the completion.
+    (list(PROGRAMS["C"][:3]), list(PROGRAMS["C"][3:])),
+]
+
+
+def test_trained_model_single_chunk_generation_recovers_grammar(grammar_trained_model: Model):
+    """For each prompt, greedy single-chunk generation must complete the in-progress program and then emit a
+    grammar-valid transition (``SEP`` or ``EOS``)."""
+    model = grammar_trained_model
+
+    for prompt, expected_completion in _PROMPTS_AND_EXPECTED_COMPLETIONS:
+        prompt_tensor = torch.tensor([prompt], dtype=torch.long)
+        batch = mock_batch(prompt_tensor)
+
+        generated = model.generate(batch, do_sample=False)  # single-chunk path
+        produced = generated[0].tolist()
+
+        # Strip any trailing PAD/EOS padding that appears after the model stopped emitting useful
+        # tokens; we only need to check enough tokens to cover the expected completion plus one
+        # more token (the transition out of the program).
+        check_len = len(expected_completion) + 1
+        assert len(produced) >= check_len, (
+            f"Expected at least {check_len} generated tokens for prompt {prompt}, "
+            f"got {len(produced)}: {produced}."
+        )
+        assert produced[: len(expected_completion)] == expected_completion, (
+            f"Greedy completion of prompt {prompt} should be {expected_completion}; "
+            f"model produced {produced[: len(expected_completion)]}."
+        )
+
+        # The token immediately after completing the program should be SEP — that's the only
+        # valid grammar transition out of the final position of a program, and since the training
+        # distribution never contains a terminating token, the model should always prefer it.
+        transition_token = produced[len(expected_completion)]
+        assert transition_token == SEP, (
+            f"After completing prompt {prompt}, expected SEP={SEP} as the transition token; "
+            f"got {transition_token}. Full output: {produced}."
+        )
+
+
+_ROLLING_BUDGET = 50
+
+
+def test_trained_model_rolling_generation_preserves_grammar(grammar_trained_model: Model):
+    """Rolling generation past the model's context window must stay grammar-valid.
+
+    The trained model has ``max_position_embeddings=16`` and we generate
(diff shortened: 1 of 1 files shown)


tests/test_pattern_generation.py
```diff
@@ -0,0 +1,383 @@
...
+    ('IN', 'A', 1)
+    >>> fsm.step(4)
+    ('IN', 'A', 2)
+    >>> fsm.step(5)
+    ('IN', 'A', 3)
+    >>> fsm.step(SEP)
+    'BETWEEN'
+    >>> fsm.step(6)  # start of program B
+    ('IN', 'B', 0)
+    >>> fsm.step(8)  # would skip B[1]=7, invalid
```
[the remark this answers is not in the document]
mmcdermott: Accepted in aab4885. The current form actually passes (the Python REPL — and doctest — suppress None results from bare expressions, so `>>> fsm.step(8)` with no expected output below is treated as "returned None, no output"), but `_ = fsm.step(8)` is strictly more explicit and avoids any doubt about what the reader/test is supposed to expect at that line. Updated to match your suggestion and added a comment so the `_ =` pattern doesn't read as arbitrary noise:

```python
>>> _ = fsm.step(8)  # would skip B[1]=7, invalid — assign to _ so doctest doesn't print None
```

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,383 @@
...
+        generated = model.generate(batch, do_sample=False, max_new_tokens=_ROLLING_BUDGET)
+    finally:
+        model.HF_model.generate = real_generate
+    produced = generated[0].tolist()
+
+    assert len(produced) == _ROLLING_BUDGET, (
+        f"Rolling generation should emit exactly max_new_tokens={_ROLLING_BUDGET} tokens (training "
+        f"data has no EOS so the loop shouldn't terminate early); got {len(produced)}. Full "
+        f"output: {produced}."
+    )
```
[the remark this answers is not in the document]
mmcdermott: You're right about the real risk, and ChatGPT flagged the same thing in a separate review. Both of you pointed at the same latent assumption — that the strict `len(produced) == _ROLLING_BUDGET` assertion depends on the model never greedy-emitting DUMMY_EOS, which is plausible but not guaranteed by any API contract. Addressed in aab4885, though I went with your second option (weaken the assertions) rather than the first (suppress the token at generation time), because it's more in the spirit of what we want to test.

**What changed:**

- Dropped `len(produced) == _ROLLING_BUDGET`.
- Dropped `spy.call_count >= 10` (which was ChatGPT's concern about coupling the test to implementation strategy).
- Replaced both with a single semantic check: `len(produced) > max_seq_len - len(prompt)`. A non-rolling `Model.generate` on the same prompt can emit at most `max_seq_len - len(prompt)` new tokens; anything longer is direct evidence that the rolling path crossed at least one sliding boundary. Combined with the existing grammar-FSM walk, this is enough to prove "rolling was exercised AND the output is grammar-valid" without caring about the exact token count, the inner call count, or whether DUMMY_EOS happened to fire early.
- Removed the `MagicMock` spy entirely since the new assertion doesn't need to count HF calls.

**Why not `suppress_tokens=[DUMMY_EOS]`:**

It would work and is a perfectly reasonable alternative, but `suppress_tokens` is an HF kwarg that passes through `**kwargs` into `HF_model.generate` — and `Model._rolling_generate` currently rejects several reserved generate kwargs in its validation check (`eos_token_id`, `pad_token_id`, `generation_config`, etc., from the PR #86 review). `suppress_tokens` isn't on that reject list, so it would work today, but any test that passes generate kwargs through the rolling loop is coupling itself to a piece of the API surface that could easily change. Keeping the test free of any HF generate kwargs means the rolling loop itself is the only thing being tested. The cost is that we trust the model not to greedy-emit DUMMY_EOS, but that trust is now cheap — the assertions no longer require it, they only require that the model emit **more than 15 non-pad tokens**, which happens within 3 or 4 program emissions regardless of what token comes next.

The test still empirically produced a full 50-token continuation across multiple rolling windows on my local CPU runs, so in practice nothing changes; only the assertion robustness changes. Thanks for the catch.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,383 @@
...
+    for prompt, expected_completion in _PROMPTS_AND_EXPECTED_COMPLETIONS:
+        prompt_tensor = torch.tensor([prompt], dtype=torch.long)
+        batch = mock_batch(prompt_tensor)
+
+        generated = model.generate(batch, do_sample=False)  # single-chunk path
+        produced = generated[0].tolist()
+
+        # Strip any trailing PAD/EOS padding that appears after the model stopped emitting useful
+        # tokens; we only need to check enough tokens to cover the expected completion plus one
+        # more token (the transition out of the program).
```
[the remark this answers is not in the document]
mmcdermott: Right — the comment promised a strip that the code never actually performed. Dropped the misleading line in aab4885 and replaced it with a comment that matches what the code actually does:

```python
# We only need to check enough tokens to cover the expected completion plus the
# transition token; anything beyond that is a don't-care here.
check_len = len(expected_completion) + 1
```

The test was only ever asserting on `produced[:len(expected_completion)]` and `produced[len(expected_completion)]`, so the "anything beyond that is a don't-care" framing is the accurate one. If a future reader wants to check the post-transition state they can add assertions without having to reason about a strip that doesn't exist.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,383 @@
...
+    return tokens
+
+
+def sample_batch(rng: random.Random, batch_size: int, max_len: int = MAX_SEQ_LEN) -> torch.Tensor:
+    seqs = [sample_sequence(rng, max_len) for _ in range(batch_size)]
+    pad_to = max(len(s) for s in seqs)
+    padded = torch.full((batch_size, pad_to), PAD, dtype=torch.long)
+    for i, s in enumerate(seqs):
+        padded[i, : len(s)] = torch.tensor(s, dtype=torch.long)
+    return padded
```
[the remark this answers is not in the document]
mmcdermott: You're right — the docstring on `sample_sequence` described a padding invariant that was actually implemented by a different function. Fixed in aab4885 by (a) clarifying the `sample_sequence` docstring to push the padding description out to `build_training_batch` and (b) adding a doctest on `build_training_batch` itself that documents the "pad to longest in batch" behavior.

Updated `sample_sequence` docstring:

> Packs programs separated by `SEP` until the next full program wouldn't fit. The sequence
> always ends with a `SEP`; there is no terminating EOS. `build_training_batch` pads a
> collection of these sequences to the longest one in the batch with `PAD`, and
> `Model._forward`'s cross-entropy loss ignores those pad positions via `ignore_index`, so
> the model never sees a "what comes after the final SEP" target and the trained grammar
> remains effectively non-terminating.

New `build_training_batch` docstring:

> Build a `[batch_size, L]` token tensor where `L` is the longest sequence in the batch.
>
> Pads shorter sequences with `PAD` on the right. Note this pads **to the longest sequence
> in the batch**, not to `max_len` — `Model._forward` ignores `PAD` via `ignore_index` in
> cross-entropy so either convention would produce the same loss, but padding only to the
> batch max saves a few forward-pass FLOPs per step.
>
> (Note: the function is named `build_training_batch` rather than `sample_batch` to avoid
> colliding with the session-scoped `sample_batch` fixture that `conftest.py` injects into
> the doctest namespace.)

Also — the function was renamed from `sample_batch` to `build_training_batch` in this same commit. That's because conftest.py's doctest namespace injects a session-scoped `sample_batch` fixture (a `MEDSTorchBatch` instance), which was shadowing my local `sample_batch` function and breaking its doctest. Fun little collision that only surfaced once I added the doctest you asked for.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,383 @@
...
+    ([PROGRAMS["C"][0]], list(PROGRAMS["C"][1:])),
+    # Mid-C: enough context to pin down the completion.
+    (list(PROGRAMS["C"][:3]), list(PROGRAMS["C"][3:])),
+]
+
+
+def test_trained_model_single_chunk_generation_recovers_grammar(grammar_trained_model: Model):
+    """For each prompt, greedy single-chunk generation must complete the in-progress program and then emit a
+    grammar-valid transition (``SEP`` or ``EOS``)."""
+    model = grammar_trained_model
```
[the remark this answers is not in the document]
mmcdermott: Fixed in aab4885 — the docstring now matches the assertion and says **only** `SEP`:

```python
def test_trained_model_single_chunk_generation_recovers_grammar(grammar_trained_model: Model):
    """For each prompt, greedy single-chunk generation must complete the in-progress program and
    then emit ``SEP`` as the transition token.

    ``SEP`` is the **only** valid transition out of the final position of a program under this
    grammar: the training distribution never contains an EOS token (see the module docstring), so
    the model cannot have learned to emit anything else at that position.
    """
```

This matches the earlier grammar redesign where I removed EOS from the training distribution to prevent the trained model from emitting it early during rolling generation (see the module docstring, *Why no EOS*). The old "SEP or EOS" wording was a stale remnant from the earlier iteration where EOS was a valid end-of-sequence token. Tightened in the two places it still lingered: this test docstring and the single-chunk test's inline comment.

mmcdermott:
## Addressed all review feedback in aab4885

Five Copilot inline comments replied inline above. Also addressing the broader ChatGPT / mmcdermott concerns here since they don't land on specific lines:

### ChatGPT's robustness concerns

1. **"Don't assert `len(produced) == _ROLLING_BUDGET` — weaken to semantic"**: Done. Replaced with `len(produced) > max_seq_len - len(prompt)` (strictly more than single-chunk capacity). Survives any rolling-loop refactor that happens to terminate at a different exact token count.
2. **"Remove `spy.call_count >= 10` — it tests implementation strategy, not contract"**: Done. Removed the spy entirely. The semantic length assertion above is sufficient to prove rolling was exercised.
3. **"Small prose/code mismatch on SEP vs EOS"**: Fixed in the `test_trained_model_single_chunk_generation_recovers_grammar` docstring and inline comment.
4. **"Runtime / flake risk"**: The new assertions are less knife-edge, which should reduce flake risk from small numerical variations across CI versus local runs. File still runs in ~54s for this file, ~100s for the full suite. Haven't added a `slow` marker — leaving that as a future option if CI actually shows pain.

### mmcdermott's concerns

1. **Developer-facing docs with runnable examples**: Added an extensive module-level doctest that shows the token vocabulary, a sample sequence, and a training batch with actual tensor contents:

   ```
   >>> rng = random.Random(0)
   >>> seq = sample_sequence(rng, max_len=16)
   >>> seq
   [9, 10, 11, 12, 13, 14, 15, 16, 17, 1]

   >>> rng = random.Random(0)
   >>> batch = build_training_batch(rng, batch_size=3)
   >>> batch
   tensor([[ 9, 10, 11, 12, 13, 14, 15, 16, 17,  1,  0,  0,  0,  0,  0],
           [ 6,  7,  8,  1,  2,  3,  4,  5,  1,  6,  7,  8,  1,  0,  0],
           [ 9, 10, 11, 12, 13, 14, 15, 16, 17,  1,  2,  3,  4,  5,  1]])
   ```

   These run under `--doctest-modules` in CI (verified locally — 1 module-level doctest + 3 per-function doctests all collected and passing). Anyone reading the file can verify the grammar encoding by inspection rather than having to trust the FSM implementation.

2. **Negative control so a buggy always-accepting FSM can't make the test vacuous**: Added in two complementary forms.

   **(a) Hand-crafted invalid sequences.** New `test_grammar_fsm_rejects_hand_crafted_invalid_sequences` feeds six specific invalid token sequences through the FSM and asserts each is rejected at the expected position. No training, fast (<1s), catches the exact always-accept failure mode.

   **(b) Cross-grammar check.** New `ALT_PROGRAMS` grammar — same tokens as `PROGRAMS`, same program start tokens, but the internal order of each program is permuted. `GrammarFSM` is now parameterized on a `programs` dict (default `PROGRAMS`), so `GrammarFSM(programs=ALT_PROGRAMS)` walks the alternative grammar. Two assertions tie this into the existing tests:

   - New `test_grammar_fsm_alt_grammar_rejects_default_grammar_sequences` proves the two grammars are materially different: each accepts its own canonical sequence and rejects the other's.
   - Extended `test_trained_model_rolling_generation_preserves_grammar` to additionally assert that `ALT_PROGRAMS`' FSM rejects the trained model's 50-token continuation within a short prefix. This proves the model learned `PROGRAMS` specifically, not a uniform-random generator, and it proves the FSM actually distinguishes correct from incorrect.

   Your exact concern — "if the checker had a bug and just always accepted, that would make this test useless and we wouldn't know, right?" — would now fail `test_grammar_fsm_rejects_hand_crafted_invalid_sequences` at the first hand-built invalid sequence. Tests wouldn't silently pass.

### Test count

The file is now 8 items:
- 1 module-level doctest (the runnable example in the module docstring)
- 3 per-function doctests (`GrammarFSM`, `sample_sequence`, `build_training_batch`)
- 2 FSM negative-control tests (`hand_crafted_invalid_sequences`, `alt_grammar_rejects_default`)
- 2 model-based integration tests (single-chunk + rolling)

55 tests in the full suite (47 from `dev` baseline + 8 in this file), ~54 seconds for this file locally and ~100 seconds for the full suite.

The branch is in a state I think is merge-ready once CI confirms green. Holding off on the explicit-weight-setting variant you mentioned in the first message — agree with you that it'd be much harder and the trained-model approach already gives us the coverage we wanted.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+            [ 6,  7,  8,  1,  2,  3,  4,  5,  1,  6,  7,  8,  1,  0,  0],
+            [ 9, 10, 11, 12, 13, 14, 15, 16, 17,  1,  2,  3,  4,  5,  1]])
+
+Row 1 is ``C | <pad>``, row 2 is ``B | A | B | <pad>``, row 3 is ``C | A`` — all valid under the
+grammar, all right-padded with ``PAD=0`` where they ran short. ``Model._forward``'s cross-entropy
+loss ignores those pad positions so the model never learns "what comes after ``PAD``."
+
+## Negative controls
+
+To protect against "always-accepting FSM" bugs that would make the generation-correctness tests
```
mmcdermott, line 72: Can you show some doctest examples of these too? These are really helpful to understand this test.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+        program-start under the active ``programs`` dict.
+      - ``("IN", prog_name, pos)`` — inside program ``prog_name`` having just emitted position
+        ``pos``. Valid next: either ``programs[prog_name][pos + 1]`` if there is one, or (at the
+        final position) ``SEP``.
+
+    No terminal state: the training distribution never emits an EOS token (see module
+    docstring), so the grammar is effectively infinite — any valid sequence of programs
+    separated by ``SEP`` can continue indefinitely.
+
+    >>> fsm = GrammarFSM()
```
mmcdermott, line 181: can you have some doctests that show the walk behavior? That's the part that really matters for these tests, right?

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+
+    Packs programs separated by ``SEP`` until the next full program wouldn't fit. The sequence
+    always ends with a ``SEP``; there is no terminating EOS. ``sample_batch`` pads a collection
+    of these sequences to the longest one in the batch with ``PAD``, and ``Model._forward``'s
+    cross-entropy loss ignores those pad positions via ``ignore_index``, so the model never sees
+    a "what comes after the final SEP" target and the trained grammar remains effectively
+    non-terminating.
+
+    >>> rng = random.Random(0)
+    >>> seq = sample_sequence(rng)
```
mmcdermott, line 307: Can you show the sequence in this doctest? It will help developers.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+    the batch**, not to ``max_len`` — ``Model._forward`` ignores ``PAD`` via ``ignore_index`` in
+    cross-entropy so either convention would produce the same loss, but padding only to the batch
+    max saves a few forward-pass FLOPs per step.
+
+    (Note: the function is named ``build_training_batch`` rather than ``sample_batch`` to avoid
+    colliding with the session-scoped ``sample_batch`` fixture that ``conftest.py`` injects into
+    the doctest namespace.)
+
+    >>> rng = random.Random(0)
+    >>> batch = build_training_batch(rng, batch_size=3)
```
mmcdermott: can you show the batch in the doctest? It will help developers.

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+            break
+        tokens.extend(prog)
+        tokens.append(SEP)
+    if not tokens:  # pragma: no cover — only triggered if max_len is absurdly small
+        tokens.extend(PROGRAMS["B"])
+        tokens.append(SEP)
+    return tokens
+
+
+def build_training_batch(rng: random.Random, batch_size: int, max_len: int = MAX_SEQ_LEN) -> torch.Tensor:
```
mmcdermott: maybe this should be called "build_training_batch_codes" as you have a mock_batch later?

tests/test_pattern_generation.py
```diff
@@ -0,0 +1,552 @@
...
+    grammar: the training distribution never contains an EOS token (see the module docstring), so
+    the model cannot have learned to emit anything else at that position.
+    """
+    model = grammar_trained_model
+
+    for prompt, expected_completion in _PROMPTS_AND_EXPECTED_COMPLETIONS:
+        prompt_tensor = torch.tensor([prompt], dtype=torch.long)
+        batch = mock_batch(prompt_tensor)
+
+        generated = model.generate(batch, do_sample=False)  # single-chunk path
```
mmcdermott, line 532: what does do_sample=False do?

mmcdermott:
can you also add another negative control -- showing that an untrained model does not pass the generation tests you have?

tests/test_pattern_generation.py
mmcdermott: Added in 0075b4d — a small doctest directly in the "Negative controls" section of the module docstring that shows `PROGRAMS` and `ALT_PROGRAMS` disagreeing concretely on a hand-written A-then-SEP sequence:

```python
>>> programs_A = [2, 3, 4, 5, 1]  # valid under PROGRAMS
>>> alt_A      = [2, 4, 3, 5, 1]  # valid under ALT_PROGRAMS
>>> GrammarFSM(programs=PROGRAMS).walk(programs_A)
5
>>> GrammarFSM(programs=ALT_PROGRAMS).walk(programs_A)
1
>>> GrammarFSM(programs=ALT_PROGRAMS).walk(alt_A)
5
>>> GrammarFSM(programs=PROGRAMS).walk(alt_A)
1
```

The `1`s are the crux: each FSM rejects the other's canonical sequence at position 1 (the first internal program transition, where A[1] differs). Added a sentence immediately after spelling that out so readers don't have to pattern-match the numbers themselves. This runs as part of the module-level doctest collection (you can see it in the verbose pytest output under `tests.test_pattern_generation`).

tests/test_pattern_generation.py
mmcdermott: You're right — `walk` is what the model tests actually call, so it should be the primary documented surface. Added doctests directly on the `walk` method in 0075b4d:

```python
def walk(self, tokens) -> int:
    """Walk a token sequence, returning the index of the first invalid token.

    Returns ``len(tokens)`` if every transition was valid (i.e. the entire sequence was
    accepted). This is the primary entry point that the model-generation tests use: after
    running ``Model.generate(...)``, they walk the resulting token list through a freshly
    initialized FSM and assert the return value equals ``len(tokens)``.

    Examples:
        A fully valid PROGRAMS-grammar sequence (program A, then SEP, then program B, then
        SEP) is accepted all the way through — ``walk`` returns the full length:

        >>> tokens = [2, 3, 4, 5, 1, 6, 7, 8, 1]  # A | B |
        >>> GrammarFSM().walk(tokens)
        9
        >>> GrammarFSM().walk(tokens) == len(tokens)
        True

        If we poison the middle of program A by swapping A[1] and A[2], ``walk`` rejects at
        the first offending position (here position 2 — A[0]=2 is fine, expected A[1]=3 but
        we emit 4 instead):

        >>> GrammarFSM().walk([2, 4, 3, 5, 1])
        1

        Emitting the wrong start token at ``BETWEEN`` is rejected at position 0:

        >>> GrammarFSM().walk([SEP])
        0
        >>> GrammarFSM().walk([99])  # not a program start
        0

        The same input is accepted by ``ALT_PROGRAMS`` (whose program A is ``(2, 4, 3, 5)``)
        but rejected by ``PROGRAMS`` (whose program A is ``(2, 3, 4, 5)``):

        >>> seq = [2, 4, 3, 5, 1]  # valid under ALT, invalid under default
        >>> GrammarFSM(programs=ALT_PROGRAMS).walk(seq)
        5
        >>> GrammarFSM(programs=PROGRAMS).walk(seq)
        1
    """
```

Covers: full-accept (returns `len(tokens)`), internal rejection (returns the index of the first bad token), BETWEEN-state rejection with both `SEP` and a random non-start token, and the cross-grammar split where `PROGRAMS` and `ALT_PROGRAMS` return different indices on the same input. I left the per-step examples on the class docstring — those document the state-machine semantics — but the `walk`-level examples are the ones that map directly onto how the generation tests use it. Pytest now collects a new doctest item `tests.test_pattern_generation.GrammarFSM.walk`.

tests/test_pattern_generation.py
mmcdermott: Done in 0075b4d. Updated the `sample_sequence` doctest to print the sequence literal (both a single seeded sample and three more consecutive draws from the same RNG so readers can see the shape variation across runs):

```python
>>> rng = random.Random(0)
>>> seq = sample_sequence(rng)
>>> seq
[9, 10, 11, 12, 13, 14, 15, 16, 17, 1]
>>> # That's program C (9..17) followed by SEP (1).
>>> GrammarFSM().walk(seq) == len(seq)
>>> seq[-1] == SEP
True

Sampling a few more seeded sequences shows how the grammar packs programs into ``max_len``:

>>> [sample_sequence(rng) for _ in range(3)]  # doctest: +NORMALIZE_WHITESPACE
[[6, 7, 8, 1, 2, 3, 4, 5, 1, 6, 7, 8, 1],
 [9, 10, 11, 12, 13, 14, 15, 16, 17, 1, 2, 3, 4, 5, 1],
 [6, 7, 8, 1, 9, 10, 11, 12, 13, 14, 15, 16, 17, 1]]
```

Now a developer opening this function in their editor can read the docstring and immediately see "ah, sequences look like program runs separated by SEP=1 with no terminating EOS" without having to go run it themselves or trust the FSM walk check.

tests/test_pattern_generation.py
mmcdermott: Done in 0075b4d — doctest now prints the full `[3, 15]` tensor with an inline walkthrough of what each row contains:

```python
>>> rng = random.Random(0)
>>> batch_codes = build_training_batch_codes(rng, batch_size=3)
>>> batch_codes
tensor([[ 9, 10, 11, 12, 13, 14, 15, 16, 17,  1,  0,  0,  0,  0,  0],
        [ 6,  7,  8,  1,  2,  3,  4,  5,  1,  6,  7,  8,  1,  0,  0],
        [ 9, 10, 11, 12, 13, 14, 15, 16, 17,  1,  2,  3,  4,  5,  1]])
```

> Row 0 is program C (9..17) then SEP (1) then PAD. Row 1 is programs B, A, B separated by SEPs then PAD. Row 2 is C then A — it fills `max_len=16` exactly, so no PAD positions. Every non-pad row is a valid program-SEP sequence under the grammar, and PAD positions at the right are ignored during training via cross-entropy's `ignore_index`.

This is the same content that's in the module-level doctest (which I added in the previous round) but now it lives where developers actually look for it — on the function that produces the batch. Dropped the property-only checks (`batch.shape[0]`, `bool(batch.shape[1] <= MAX_SEQ_LEN)`) since the printed tensor makes them self-evident.

tests/test_pattern_generation.py
mmcdermott: Good call — renamed in 0075b4d to `build_training_batch_codes`. The `_codes` suffix does two useful things at once:

1. **It disambiguates from `mock_batch`.** `mock_batch(codes)` wraps a codes tensor in a minimal `MEDSTorchBatch`-shaped mock so `Model.forward` / `Model.generate` can consume it. The two-step split (`build_training_batch_codes` → `mock_batch`) keeps training noise-free: the training loop only needs the codes tensor, and calling `mock_batch` inside the loop would be pure ceremony.
2. **It makes the return type self-documenting.** "Batch" in this codebase generally means `MEDSTorchBatch`, not a raw tensor — which is exactly the confusion you were pointing at. `build_training_batch_codes` reads as "builds the `.code` tensor that goes into a batch" without any ambiguity.

As a bonus, this also sidesteps the conftest-fixture shadowing that I hit with the earlier `sample_batch` name, though that was already resolved in the previous round's rename to `build_training_batch`. Now the name simultaneously reflects the return type, disambiguates from `mock_batch`, and avoids the fixture collision.

Updated all three call sites (function definition, module-level doctest, training-loop call inside `grammar_trained_model`).

tests/test_pattern_generation.py
mmcdermott: Fair — added an explanation in both the single-chunk and rolling test docstrings in 0075b4d.

**Single-chunk test docstring:**

> Generation uses `do_sample=False` so the model greedy-decodes each next token (argmax over logits, no temperature sampling). That makes the output fully deterministic given the trained weights and the prompt, so a one-shot assertion against the expected program completion is meaningful — we're testing "does the model's argmax follow the grammar?", not "does some stochastic sample happen to do so."

**Rolling test inline comment at the call site:**

```python
# do_sample=False → greedy argmax decoding, same as the single-chunk test: the output is
# fully determined by the trained weights and the prompt, so we can assert on specific
# grammar-FSM transitions rather than on distributions.
generated = model.generate(batch, do_sample=False, max_new_tokens=_ROLLING_BUDGET)
```

**Why this matters for the test design.** `do_sample=False` is HF's way of saying "greedy decoding" — at each step the model emits `argmax(logits)` rather than sampling from `softmax(logits)`. Two reasons we want it here:

1. **Determinism.** If the model is well-trained, argmax is fully determined by the prompt. That means the test can make deterministic assertions like "after emitting `A[0]`, the next token must be `A[1]`" and catch regressions exactly — without having to reason about sampling temperature or random seeds affecting the output.
2. **Strongest test of learning.** Greedy decoding tests whether the model has *actually* learned the grammar, not whether it has some non-trivial probability on the correct token. A model that learned a perfect grammar will argmax to the correct next token with probability 1 after enough training, so greedy tests the most stringent case.

For generation tests that care about coverage or diversity we'd want sampling, but for correctness-of-learned-grammar we want greedy. Same reason the model.py doctests use `do_sample=False`.

PR #94, opened 2026-04-15
Adding rolling generation and a better generation test.

mmcdermott:


mmcdermott:
## Pre-release review of dev → main

I went through the full `git diff main..dev` (13 non-merge commits, ~1500 added lines across 9 files) and ran the full test suite locally on dev. **Verdict: merge-ready. No blockers. Two low-severity docstring/comment nits that can be cleaned up in a follow-up or now, your call.**

### What's in this release

- **Rolling (sliding-window) generation** (`Model.generate` + `Model._rolling_generate` + `Model._generate_chunk`) that lets the generation path emit more tokens than `max_position_embeddings`. Shared `_generate_chunk` helper between the single-chunk and rolling paths. Validation of reserved HF kwargs. Hoisted config validation in the CLI.
- **`generation/utils.py::validate_rolling_cfg`** — OmegaConf-based config validator with 11 doctests covering every failure mode.
- **Auto-populated `eos_token_id`** in `MEICAR_generate_trajectories` so older checkpoints work without manual config intervention.
- **CI coverage fix** (PR #92) — `--cov-report=xml` and explicit `files: ./coverage.xml` so codecov stops reporting "unknown".
- **`tests/test_pattern_generation.py`** (PR #93) — a real generation-correctness integration test: tiny model trained on a synthetic finite-state grammar, verified via grammar-FSM walk of the generated tokens. Covers both single-chunk and rolling, with `ALT_PROGRAMS` negative control to prove the FSM isn't always-accepting.
- **`tests/test_generate_trajectories.py`** — extended with CLI-level rolling integration test and three direct unit tests on `Model._rolling_generate`.

### What I checked and passed

- **Full test suite on dev:** 56/56 green locally (`uv run pytest -q`, ~44s).
- **CI on PR #94:** `run_tests_ubuntu` pass (2m26s), `code-quality` pass (1m23s), build pass. No failing checks, `mergeStateStatus=CLEAN`.
- **No stub functions, no commented-out imports, no half-finished TODOs added.** The only new `TODO(generation_test)` was already present on main and is now arguably satisfied by `tests/test_pattern_generation.py`.
- **Single-chunk path refactor is equivalence-preserving.** The old inlined `self.HF_model.generate(for_hf.pop("input_ids"), generation_config=..., **for_hf, **kwargs)` now goes through `_generate_chunk(input_ids, attention_mask=for_hf.get("attention_mask"), ..., **kwargs)`. `_hf_inputs` only returns `input_ids` and `attention_mask`, so both call patterns forward the same effective kwargs.
- **Cross-chunk `finished` mask logic is correct.** When a row hits EOS mid-chunk, HF pads the rest of that chunk's output via `pad_token_id`. When rolling constructs chunk k+1's prompt, the visible tail contains the EOS token, and HF would happily keep generating — but the `masked_fill(finished, pad_id)` runs *before* new tokens are written into `sequence_so_far`, so the row stays padded. `finished = finished | hit_eos` persists the flag across iterations; because `eos_id != pad_id` is validated, masked-to-pad rows never re-trigger `hit_eos`.
- **Validation in `_rolling_generate` is complete.** Reserved-kwargs check, `max_new_tokens > 0`, `rolling_context_size > 0`, `eos_id is not None`, `eos_id != pad_id`. All have doctests.
- **`validate_rolling_cfg` edge cases.** Accepts `None`/all-None configs (rolling disabled); rejects unknown keys, non-ints, bools, non-positive ints, and `rolling_context_size` without `max_new_tokens`. Explicitly rejects bool (since `bool ⊂ int` in Python). 11 doctests.
- **Version bump:** `pyproject.toml` uses `dynamic = ["version"]` via `setuptools_scm`, so the next release will auto-tag from git — no manual version bump needed. Current latest tag is `0.1.11`; this release would cut `0.1.12` on merge to main.
- **No changes to public API surface** beyond the two new kwargs on `Model.generate` (`max_new_tokens`, `rolling_context_size`), which default to `None` and preserve the legacy path.

### Nits (optional, not blockers)

1. **Stale docstring reference in `_rolling_generate`** at `src/MEDS_EIC_AR/model/model.py:672`. The `Raises:` section mentions "desynchronize HF's in-chunk stopping from this function's cross-chunk `finished` mask and post-loop EOS truncation" — but the `_truncate_at_first_eos` helper was deleted in 7c26116 (sequence_so_far refactor). The "post-loop EOS truncation" wording should be dropped. Same issue in the `Returns:` section at line 662-663 which says "Each row is truncated at its first EOS, with post-EOS positions replaced with `PAD_INDEX`" — effect is still correct (HF's in-chunk padding + the finished mask), but the "truncated" wording implies an explicit pass that no longer exists.

2. **Config comment slight mismatch** in `src/MEDS_EIC_AR/configs/_generate_trajectories.yaml`. The comment says `rolling_context_size: null defaults to pretrained_max_seq_len - 1`, which is true — but the actual default computation lives in `_rolling_generate` (`ctx_size = self.max_seq_len - 1` when `rolling_context_size is None`), not in the config. Reading the YAML file doesn't make it clear where the default gets applied. Minor.

Both are pure documentation touch-ups — no behavior change. I can push a fix commit if you want them cleaned up before merge, or defer to a follow-up. Your call.

### Recommendation

**Merge.** Neither nit is a correctness issue, and the 1500-line diff is otherwise clean, well-tested, and consistent with the rest of the codebase. Cut the `0.1.12` release from main after merge.

Issue #97, opened 2026-04-16
vLLM backend integration as alternative/fallback to SGLang


mmcdermott:
## Summary

Integrate vLLM as an optional generation backend, as a lower-risk alternative to SGLang (#88). vLLM has clear documentation for offline inference with token-ID prompts and a documented offline custom-logits-processor API. GPTNeoX is natively registered in vLLM's model registry.

## Why a separate issue from SGLang

SGLang is preferred if SCOPE/REACH integration is on the roadmap (the reference implementation runs on SGLang, and RadixAttention is ideal for N-completions-per-subject workloads). But vLLM is worth having as a documented fallback because:

- GPTNeoX support in SGLang hasn't been explicitly verified yet (spike item in #88).
- vLLM's offline engine API and custom logits processor support are more thoroughly documented for offline (non-server) use.
- If SGLang turns out to have compatibility issues, having a vLLM path means we're not blocked.

## Scope

Same Option C architecture as #88: swap `Model._generate_chunk` inner call for `vllm.LLM.generate(prompt_token_ids=..., sampling_params=...)`. Rolling-loop bookkeeping stays in our code.

Key items:
1. Weight export from Lightning checkpoint to HF `save_pretrained` format.
2. `vllm.LLM(model=..., skip_tokenizer_init=True)` to bypass BPE tokenization.
3. `SamplingParams(max_tokens=chunk_budget, stop_token_ids=[eos_id])` for per-chunk generation.
4. Prefix caching via `enable_prefix_caching=True` for N-trajectories-per-subject interleaving (#89).

## Dependencies

- #89 (interleaving) for prefix-cache benefit.
- #82 (logits-processor plumbing) for time-budget / REACH processors.
- This is a parallel/alternative path to #88 (SGLang), not a prerequisite or dependency.

## References

- [vLLM offline inference](https://docs.vllm.ai/en/latest/serving/offline_inference/)
- [vLLM prefix caching](https://docs.vllm.ai/design/prefix_caching.html)

mmcdermott:
## Investigation: vLLM fit for this project + should we support both vLLM and SGLang?

Ran this as part of the engine-backend investigation after discovering SGLang doesn't support `GPTNeoXForCausalLM` (see #88 comment). Short version: **vLLM is a clean fit and covers our needs; a single-backend strategy beats dual-backend for a research codebase**.

### vLLM-specific gotchas for this project

[...]

Issue #88, opened 2026-04-14
Concrete SGLang backend setup for MEDS-EIC-AR (complements #87)


mmcdermott:
## Context

#87 establishes the *strategic* case for adopting SGLang as an optional high-throughput generation backend (PagedAttention / RadixAttention, continuous batching, prefill-only scoring, and concrete compatibility with the SCOPE/REACH variance-reduction research — see [arXiv 2602.03730](https://arxiv.org/abs/2602.03730) and https://github.com/lukesolo-ml/SCOPE_REACH_optimized_inference). What #87 doesn't yet contain is the concrete walkthrough of what the integration actually looks like in this repo. This issue is that walkthrough, scoped so it can be picked up and implemented as a single piece of work once the prerequisites from #82, #86, and the #87 comment thread land.

This issue assumes we've already done (or are doing in parallel):

- Persistent sliding KV cache in `_rolling_generate` (follow-up to #86, captured in #87 §Option D).
- Generic logits-processor plumbing in `Model.generate` that persists processor state across rolling chunks (requirement from the #87 SCOPE/REACH comment thread).

With those pieces in place, this issue covers everything else needed to route generation through SGLang as an optional backend, behind a config flag, with the HF path remaining the default and fully functional.

## 1. Dependency

SGLang is a **strictly optional extra**. It pulls flash-attn / triton / CUDA-heavy trees; we don't want `pip install MEDS-EIC-AR` to drag them in by default.

```toml
# pyproject.toml
[project.optional-dependencies]
sglang = ["sglang>=0.4", "torch>=2.4"]
```

The import is lazy: `import sglang as sgl` only happens inside the new backend module, never at top-level. Users without the extra see no change, and CPU-only CI lanes don't attempt to resolve it.

## 2. Lightning checkpoint → HF on-disk export

SGLang loads models from an HF on-disk directory (`config.json`, `*.safetensors`, etc.), not from a Lightning `.ckpt`. Our checkpoints wrap a real `GPTNeoXForCausalLM`, so the export is a one-liner, cached next to the checkpoint:

```python
# src/MEDS_EIC_AR/model/export.py  (new)
def export_to_hf_dir(module: "MEICARModule", out_dir: Path) -> Path:
    """Materialize a Lightning checkpoint as an HF on-disk model directory."""
    out_dir.mkdir(parents=True, exist_ok=True)
    module.model.HF_model.save_pretrained(out_dir, safe_serialization=True)
[...]

Issue #87, opened 2026-04-14
Fast generation backend: vLLM / SGLang integration + sliding KV cache (relates to #82, #86)


mmcdermott:
## Summary

Current generation goes through `HF_model.generate()` — the slowest path available for autoregressive inference. Several production-grade frameworks (vLLM, SGLang) offer 5–20× throughput improvements through PagedAttention, continuous batching, prefix caching, speculative decoding, and quantization. At least one of them (**SGLang**) is already a dependency of downstream trajectory-evaluation research we are likely to adopt (SCOPE/REACH, see references below), which makes the backend choice strategic as well as performance-driven.

This issue scopes out the options so we can pick a direction. It relates to but is distinct from:

- **#82** — rolling generation with EOS / time-budget stopping (the *algorithm*; this issue is about the *backend*).
- **#86** — just-landed rolling sliding-window loop; every chunk currently re-prefills the KV cache, which is the biggest single inefficiency we control today.
- **#53** — generation speed logging; any backend change should leave us able to measure tokens/sec.
- **`MEDS_trajectory_evaluation`** — some of what SCOPE/REACH delivers lives in trajectory evaluation, not here, so we should coordinate rather than duplicate.

## Current state

- `Model.generate()` calls `HF_model.generate()` (one-shot path) or, since #86, loops HF generate across sliding-context chunks (`_rolling_generate`).
- Each chunk in `_rolling_generate` throws away its KV cache and re-prefills from scratch — that's O(T · ctx²) work where T is total generated length.
- Flash Attention 2 is wired in conditionally (`model.py:19-26, 162-178`).
- No prefix reuse across batches, no continuous batching, no quantization, no speculative decoding, no persistent KV.

## References

- **vLLM** — https://github.com/vllm-project/vllm
- **SGLang** — https://docs.sglang.io/index.html
- **SCOPE / REACH (variance-reduced Monte Carlo for event estimation from AR trajectories)** — https://github.com/lukesolo-ml/SCOPE_REACH_optimized_inference
- **cocoa (bbj-lab)** — https://github.com/bbj-lab/cocoa — preprocessing pipeline, listed for context; **not** a generation speedup (it's a MEDS-format data collation/tokenization/winnowing pipeline). Mentioned here only so future readers don't mistake it for an inference optimization.

### Option A — Integrate vLLM as an optional backend

[...]

PR #103, opened 2026-04-16
feat: interleave N samples per subject into one predict pass (closes #89)

### LINKED ISSUE #89: Restructure generate_trajectories: batch N samples per subject into single pass
## Summary

When generating `N_trajectories_per_task_sample` samples per subject, `MEICAR_generate_trajectories` currently runs a **full pass over the dataloader once per sample** (`src/MEDS_EIC_AR/__main__.py:143-172`). If you're generating 20 samples per subject across 3 splits, that's 60 full predict passes, each of which re-prefills every subject's input from scratch and never shares work between passes.

A much better shape — on any backend, including the current HF one — is to **do one pass, with each subject contributing `N` rows to the batches instead of one**. Same-subject rows end up adjacent in batches, which gives us three distinct wins:

1. **On prefix-caching backends (vLLM / SGLang, see #87):** the KV cache for a subject's input is built once and reused across all `N` samples of that subject in the same batch, instead of being re-prefilled `N` times across `N` separate passes. This is the single biggest generation-time speedup available to us once we have such a backend — potentially `N×` on the decode-heavy path.
2. **On any backend including HF:** when a batch contains many copies of the same subject (or same-length subjects), **padding tightens dramatically**. In the limit where `batch_size == N`, every row in a batch is the same length and there is zero padding waste. Current behavior batches heterogeneous subjects and pads to the max-length in each batch, so short subjects pay a padding overhead that grows with the length-variance of the sample.
3. **One iteration of per-pass setup instead of N.** Lightning `trainer.predict` initialization, dataloader worker spawning, GPU warmup, hash-seed reseeding — all of these happen 60 times today on a 3-split × 20-sample run. Amortized, this isn't enormous, but it's non-zero, and it's entirely avoidable.

This issue is **independent of #87**: it's pure work on the CLI / data-layer and doesn't require any backend changes. It's worth doing on its own for the padding win (2) and the setup-amortization win (3), and it's a **prerequisite multiplier** for the SGLang backend work in #87 because without it most of SGLang's prefix-cache advantage is wasted.

## Current state

From `src/MEDS_EIC_AR/__main__.py:143-172`:

```python
for split in inference.generate_for_splits:
    dataloader = ...  # full dataset for this split

    for sample in range(inference.N_trajectories_per_task_sample):
        out_fp = Path(cfg.output_dir) / split / f"{sample}.parquet"
        ...
        seed = hash_based_seed(cfg.get("seed", None), split, sample)
        seed_everything(seed, workers=True)
        predictions = trainer.predict(model=M, dataloaders=dataloader)    # ← full pass
        predictions_df = format_trajectories(dataloader.dataset, predictions)
        pq.write_table(...)
```

Each `trainer.predict(...)` call iterates the entire dataset. For `N = 20` samples per subject, **every subject's input tensor is materialized and prefilled 20 times** — once per outer-loop iteration — with no sharing between iterations.

### High-level

Replace the nested `for sample in range(N): trainer.predict(...)` with a **single predict pass** over a dataset that's been expanded to length `N × len(base_dataset)`, with each base item contributing `N` consecutive rows. Same-subject rows are guaranteed adjacent, so the dataloader's batches contain tight groups of same-subject copies.

Per-row metadata — `(subject_idx, sample_idx)` — rides through the batch so that `format_trajectories` and the parquet writer can unscramble the flat predictions back into the per-sample output files.

### Dataset wrapper

```python
# src/MEDS_EIC_AR/generation/repeated_dataset.py  (new)
class RepeatedPredictionDataset(torch.utils.data.Dataset):
    """Wraps a base dataset so each underlying item contributes `n_samples` consecutive rows.

    The ordering is (subject_idx changes slow, sample_idx changes fast), so a batch of
    ``batch_size`` covers at most ``ceil(batch_size / n_samples)`` distinct subjects and at least
    ``floor(batch_size / n_samples)`` full sample-groups.

    Concretely: if n_samples=4 and the base dataset is [A, B, C, ...], this dataset yields
        A#0 A#1 A#2 A#3 B#0 B#1 B#2 B#3 C#0 ...
    with each `A#k` carrying metadata ``(subject_idx=0, sample_idx=k)``.
    """

    def __init__(self, base: Dataset, n_samples: int):
        if n_samples < 1:
            raise ValueError(f"n_samples must be >= 1, got {n_samples}")
        self.base = base
        self.n_samples = n_samples

    def __len__(self) -> int:
        return len(self.base) * self.n_samples

    def __getitem__(self, idx: int):
        subject_idx, sample_idx = divmod(idx, self.n_samples)
        item = self.base[subject_idx]
        # The base item is an opaque object (MEDSTorchBatch-compatible dict or dataclass).
        # We attach metadata as a sidecar; the collate_fn repacks it onto the batch.
        return item, subject_idx, sample_idx
```

### Collate

The base `MEDSPytorchDataset.collate` already knows how to collate its native items. We wrap it:

```python
def collate_with_meta(raw_items, base_collate):
    items, subject_idxs, sample_idxs = zip(*raw_items)
    batch = base_collate(list(items))
    # MEDSTorchBatch is a dataclass; extend with sidecar tensors for the predict path.
    batch.subject_idxs = torch.as_tensor(subject_idxs, dtype=torch.long)
    batch.sample_idxs = torch.as_tensor(sample_idxs, dtype=torch.long)
    return batch
```

If `MEDSTorchBatch` doesn't accept extra attributes cleanly, carry the metadata via a simple `(batch, subject_idxs, sample_idxs)` tuple and unpack at `predict_step`.

### Module / CLI changes

```python
# src/MEDS_EIC_AR/training/module.py
class MEICARModule(L.LightningModule):
    ...
    def predict_step(self, batch, batch_idx: int, dataloader_idx: int = 0):
        # Per-batch seeding. Reproducibility is now at batch granularity rather than
        # per-sample — see §Seeding tradeoffs below.
        if self._predict_seed_fn is not None:
            seed_everything(self._predict_seed_fn(batch_idx), workers=False)
        tokens = self.model.generate(batch, **self.generation_kwargs)
        return {
            "tokens": tokens,                         # [B, T]
            "subject_idxs": batch.subject_idxs,       # [B]
            "sample_idxs": batch.sample_idxs,         # [B]
        }
```

```python
# src/MEDS_EIC_AR/__main__.py::generate_trajectories
for split in inference.generate_for_splits:
    base_loader = ...
    base_dataset = base_loader.dataset
    N = inference.N_trajectories_per_task_sample

    expanded = RepeatedPredictionDataset(base_dataset, N)
    loader = DataLoader(
        expanded,
        batch_size=base_loader.batch_size,
        shuffle=False,
        num_workers=base_loader.num_workers,
        collate_fn=partial(collate_with_meta, base_collate=base_dataset.collate),
        pin_memory=base_loader.pin_memory,
    )

    M._predict_seed_fn = lambda batch_idx, s=split: hash_based_seed(
        cfg.get("seed", None), s, batch_idx
    )

    predictions = trainer.predict(model=M, dataloaders=loader)

    # Reshape flat [batch] predictions back into per-sample_idx groups and write
    # one parquet per sample.
    by_sample: dict[int, list] = defaultdict(list)
    for pred in predictions:
        for i in range(pred["tokens"].shape[0]):
            by_sample[int(pred["sample_idxs"][i])].append({
                "subject_idx": int(pred["subject_idxs"][i]),
                "tokens": pred["tokens"][i],
            })

    for sample_idx, rows in by_sample.items():
        out_fp = Path(cfg.output_dir) / split / f"{sample_idx}.parquet"
        out_fp.parent.mkdir(parents=True, exist_ok=True)
        if out_fp.is_file() and not cfg.do_overwrite:
            continue
        predictions_df = format_trajectories_from_rows(base_dataset, rows)
        pa_table = GeneratedTrajectorySchema.align(predictions_df.to_arrow())
        pq.write_table(pa_table, out_fp)
```

`format_trajectories_from_rows` is a small shim that adapts the existing `format_trajectories` to the new flat-row shape; in practice it may just be a loop into the existing function with the `(subject_idx, tokens)` pairs for one sample.

## The padding argument, quantified

Given a batch of size `B` where each row has length `L_i`:

- **Current behavior (heterogeneous batches):** pad-to-max per batch. Wasted padding tokens per batch ≈ `B · max_i(L_i) - sum_i(L_i)`. If subject lengths vary significantly within the split (which is normal — some patients have short histories, some long), this waste can approach 50%+ of batch compute on short rows.
- **With sample-grouped batches and `batch_size = k · N` for integer `k`:** each batch contains `k` distinct subjects × `N` copies. The `N` copies are identical in length, so waste is `B · max_i(L_i) - sum_{k distinct}(N · L_i)`. When `k = 1` (`batch_size = N`), waste is **exactly zero**: every row in the batch has the same length.
- **Even without tuning `batch_size` to a multiple of `N`:** adjacent-subject clustering still reduces length variance within each batch (you're averaging over fewer distinct subjects), so max-length-per-batch tightens.

For workloads where `N_trajectories_per_task_sample` is `>= 8` (research configs) and subject lengths are long-tailed (realistic), this is a meaningful GPU-utilization win even on HF with no other changes.

## Prefix cache argument (for #87)

On a prefix-caching backend (vLLM / SGLang), same-subject rows within a batch share KV-cache entries for the prompt prefix. Generation of `N` samples from one subject therefore:

- Pays the prompt-prefill cost once, not `N` times.
- Pays the decode cost `N` times (unavoidable, samples diverge after the first sampled token).

On a full-size model where prefill dominates decode, this is approximately a **`(1 + (N-1)/N) × speedup ≈ N×`** reduction in prefill cost, or equivalently a large fraction of total generation cost.

**None of this win is accessible under the current looped-predict design**, because each pass creates a fresh batch and the cross-pass cache reuse that SGLang's radix tree could in principle offer is blocked by the outer loop's batch reshuffling. The batching-order change in this issue unblocks it.

## Seeding tradeoffs

The current design gives **per-sample reproducibility**: fixing `cfg.seed` and `(split, sample)` produces the same trajectory for a given subject every run.

The proposed design gives **per-batch reproducibility**: fixing `cfg.seed`, the split, and the batch index produces the same batch of trajectories. Row ordering within a batch is deterministic, so the per-(subject, sample) mapping is also stable *given* the batch layout — but it depends on `batch_size` and dataset ordering in a way the current design doesn't.

Two mitigations, both optional:

- **Mitigation A (sufficient for research use):** document the new contract. Seeds are reproducible as long as `batch_size`, `N_trajectories_per_task_sample`, and the dataset itself are unchanged. If any of those change, the same `(subject, sample_idx)` pair may map to a different trajectory. This is usually fine.
- **Mitigation B (stronger, more work):** use backend-level per-request seeds. vLLM and SGLang both accept `SamplingParams(seed=...)` per request, so we can compute `hash_based_seed(cfg.seed, split, subject_idx, sample_idx)` per row and pass it through. HF does not cleanly support per-row seeding in a single `generate()` call, so on the HF backend Mitigation B would require grouping rows by seed and running multiple sub-batches — negating most of the padding win. Recommendation: adopt Mitigation A as the default, and add Mitigation B as an opt-in flag only for the non-HF backends once #87 lands.

## Interaction with existing issues

- **Independent of #87.** No backend changes required. HF-only users still benefit (padding and setup amortization).
- **Prerequisite multiplier for #87.** Without this change, SGLang / vLLM can't reuse KV across samples of the same subject, so the prefix-cache win in #87 is mostly vestigial.
- **Compatible with #86 rolling generation.** The rolling loop runs per-batch-row independently; grouping N same-subject rows into one batch changes nothing about the rolling logic. It just means the rolling loop processes `k · N` rows per batch instead of `k` rows across `N` separate passes.
- **Neutral to #82 time-budget stopping.** The time budget is per-row state; grouping affects throughput, not the stopping condition.
- **Neutral to #53 generation speed logging.** If anything, this makes speed logging more meaningful because tokens-per-second is no longer diluted by the outer-loop overhead.

## Tests

- **Unit test on `RepeatedPredictionDataset`:** length, indexing, metadata attachment. Verify `(subject_idx, sample_idx)` round-trip.
- **Unit test on `collate_with_meta`:** sidecar tensors present and correctly shaped.
- **Integration test (new end-to-end):** replace the existing `generated_trajectories` fixture's CLI invocation with the new code path. Assert the same output structure: one parquet per `(split, sample_idx)` with the right subjects and non-empty data. Assert that the test passes with `N_trajectories_per_task_sample > 1` — which is where the restructuring actually matters.
- **Parity smoke test:** run the new path with `N=1`; it should produce output shape-identical to the old single-pass loop. (Exact token parity is not guaranteed because the seeding granularity changes, but structural parity is.)
- **Padding assertion (optional):** construct a batch with variable-length subjects and verify that with the grouped path, within-batch max-length variance is ≤ the per-subject length variance. This documents the padding win.

## Implementation checklist

- [ ] `src/MEDS_EIC_AR/generation/repeated_dataset.py` with `RepeatedPredictionDataset` and `collate_with_meta`.
- [ ] `MEICARModule.predict_step` returns `{"tokens", "subject_idxs", "sample_idxs"}` and accepts a `_predict_seed_fn`.
- [ ] `__main__.py::generate_trajectories` replaces the `for sample in range(N)` loop with one predict pass over the expanded dataset, then reshapes into per-sample parquets.
- [ ] `format_trajectories` adapted or wrapped to accept the new flat-row shape.
- [ ] Unit tests for the dataset + collate.
- [ ] End-to-end test mirroring the existing `test_generate_trajectories_runs` but using the new path.
- [ ] Documentation note in `configs/README.md` explaining the new seeding contract.

## Motivation recap

The TL;DR of why this is worth doing even before any backend work:

1. **Padding efficiency.** Tightly-grouped batches → near-zero padding waste → higher effective GPU utilization per batch → faster wall-clock even on the current HF backend.
2. **Setup amortization.** One predict pass instead of `N`; one round of worker spawning, one round of dataloader warmup, one round of seed-everything boilerplate.
3. **Unblocks prefix caching.** This is the change that turns #87's SGLang / vLLM adoption from "nice kernel speedup" into "`N×` generation throughput on research configs."
4. **Better speed telemetry.** #53's speed logging becomes more honest because the outer-loop overhead isn't drowning the true decode throughput.
5. **Zero behavioral regression for `N=1` use cases.** The change degenerates into the current path when there's only one sample per subject.

## Labels
`enhancement`, `priority:high`, `Computational Performance`, `Trajectory Generation / Inference`

mmcdermott:
## Summary

Closes #89. Replaces the `for sample in range(N): trainer.predict(...)` loop in `MEICAR_generate_trajectories` with **one predict pass over an expanded dataset** where each base subject contributes `N` consecutive rows.

Wins on any backend, multiplied on prefix-caching backends:

1. **Tighter padding.** Within each batch, same-subject rows have identical length, so the per-batch max-length tightens vs. the heterogeneous batches the old loop produced.
2. **Setup amortization.** Lightning init / dataloader worker spawn / GPU warmup happen once per split instead of N times.
3. **Prefix-cache reuse on backends that have one** (vLLM/SGLang per #88, #97). The KV cache for a subject's prompt prefix is built once and reused across all N samples of that subject in the same batch, instead of being re-prefilled across N separate passes. **Dormant on the current HF backend; activates automatically when SGLang lands.**

## Implementation

**New module `src/MEDS_EIC_AR/generation/repeated_dataset.py`:**
- `RepeatedPredictionDataset(base, n_samples)` — wraps the base dataset with subject-changes-slow / sample-changes-fast ordering.
- `collate_with_meta(items, base_collate)` — runs the base collate and attaches `(subject_idxs, sample_idxs)` per-row tensors via a sidecar attribute.
- `extract_meta(batch)` — reads the metadata back with an actionable error if the batch wasn't routed through `collate_with_meta`.

**`MEICARModule.predict_step`** now returns `{tokens, subject_idxs, sample_idxs}` when metadata is present (legacy tokens-only return preserved when not, so direct `Model.generate` callers are unaffected).

**`MEICAR_generate_trajectories`** builds the expanded DataLoader, runs one predict pass per split, then demuxes the flat predictions into per-sample batch lists and feeds each to `format_trajectories` for the existing per-sample parquet write path. Idempotent partial-skip support preserved.

**`hash_based_seed`** gets a `sample=0` default so the new single-seed-per-split call site (`hash_based_seed(seed, split)`) works without breaking the per-sample-seeded API for any other caller.

## Tests

- `RepeatedPredictionDataset` doctest covers length, indexing, metadata, validation.
- `test_collate_with_meta_round_trip_through_dataloader` runs the full pipeline through a real DataLoader on a fake dataset and confirms the per-row metadata round-trips.
- `test_extract_meta_raises_on_unwrapped_batch` verifies the error path.
- The existing CLI integration tests (`test_generate_trajectories_runs`, `test_generate_trajectories_rolling_runs`) exercise the new path end-to-end with N=2 from the demo inference config; both pass.
- 59/59 tests pass locally.

## What's next

Per the roadmap restructuring in #87 (now closed): with #89 landed, **#88 (SGLang integration)** unlocks the prefix-cache benefit. That's the next planned PR.

### DIFF
--- a/src/MEDS_EIC_AR/__main__.py
+++ b/src/MEDS_EIC_AR/__main__.py
@@ -2,6 +2,7 @@ import logging
 import os
 import shutil
 from datetime import UTC, datetime
+from functools import partial
 from importlib.resources import files
 from pathlib import Path
 
@@ -15,8 +16,15 @@ from meds_torchdata import MEDSTorchBatch, MEDSTorchDataConfig
 from MEDS_trajectory_evaluation.schema import GeneratedTrajectorySchema
 from MEDS_transforms.runner import load_yaml_file
 from omegaconf import DictConfig, OmegaConf
+from torch.utils.data import DataLoader
 
-from .generation import format_trajectories, get_timeline_end_token_idx, validate_rolling_cfg
+from .generation import (
+    RepeatedPredictionDataset,
+    collate_with_meta,
+    format_trajectories,
+    get_timeline_end_token_idx,
+    validate_rolling_cfg,
+)
 from .training import MEICARModule, find_checkpoint_path, validate_resume_directory
 
 # Import OmegaConf Resolvers
@@ -180,34 +188,79 @@ def generate_trajectories(cfg: DictConfig):
     if cfg.get("seed", None):
         seed_everything(cfg.get("seed", 1), workers=True)
 
+    n_samples = inference.N_trajectories_per_task_sample
+
     for split in inference.generate_for_splits:
         if split == train_split:
-            dataloader = D.train_dataloader()
+            base_loader = D.train_dataloader()
         elif split == tuning_split:
-            dataloader = D.val_dataloader()
+            base_loader = D.val_dataloader()
         elif split == held_out_split:
-            dataloader = D.test_dataloader()
+            base_loader = D.test_dataloader()
         else:
             raise ValueError(f"Unknown split {split}.")
 
-        for sample in range(inference.N_trajectories_per_task_sample):
-            out_fp = Path(cfg.output_dir) / split / f"{sample}.parquet"
-            out_fp.parent.mkdir(parents=True, exist_ok=True)
+        # Skip work for samples whose output parquet already exists. If every requested sample is
+        # already on disk and ``do_overwrite`` is false, skip the predict pass entirely; otherwise
+        # we still run a single pass over the full ``N``-expanded dataset and just don't write the
+        # parquets that already exist. (Partial-skip support is a minor wrinkle — it keeps existing
+        # checkpointed runs idempotent without making us special-case mid-run resumption.)
+        sample_paths = {
+            sample: Path(cfg.output_dir) / split / f"{sample}.parquet" for sample in range(n_samples)
+        }
+        for sample_fp in sample_paths.values():
+            sample_fp.parent.mkdir(parents=True, exist_ok=True)
+        if not cfg.do_overwrite and all(p.is_file() for p in sample_paths.values()):
+            logger.info(f"Skipping all {n_samples} samples for split {split}: every parquet exists.")
+            continue
 
+        # Expand the base dataset so each subject contributes ``n_samples`` consecutive rows. See
+        # issue #89 for the motivation: one predict pass instead of ``N``, tighter padding, and
+        # prefix-cache reuse on backends that have one (#88, #97). The ordering invariant — subject
+        # changes slow, sample changes fast — means rows for sample ``s`` extracted from each batch
+        # in order land in subject-index order overall, which is what ``format_trajectories`` needs
+        # so its sequential ``schema_df.slice(...)`` lines up with the right subject metadata.
+        base_dataset = base_loader.dataset
+        expanded_dataset = RepeatedPredictionDataset(base_dataset, n_samples=n_samples)
+        expanded_loader = DataLoader(
+            expanded_dataset,
+            batch_size=base_loader.batch_size,
+            shuffle=False,
+            num_workers=base_loader.num_workers,
+            collate_fn=partial(collate_with_meta, base_collate=base_dataset.collate),
+            pin_memory=base_loader.pin_memory,
+        )
+
+        seed = hash_based_seed(cfg.get("seed", None), split)
+        logger.info(
+            f"Generating {n_samples} trajectories for each of {len(base_dataset)} subjects in split "
+            f"{split} (one interleaved predict pass over {len(expanded_dataset)} expanded rows, "
+            f"seed={seed})."
+        )
+        seed_everything(seed, workers=True)
+        predictions = trainer.predict(model=M, dataloaders=expanded_loader)
+
+        # Demux the flat predictions into per-sample, per-batch token lists. Within each batch the
+        # rows for sample ``s`` are in subject-index order (because the expanded dataset was built
+        # with subject-changes-slow ordering and ``shuffle=False``), and across batches the
+        # subject-index ranges are non-overlapping and increasing — so the concatenation per sample
+        # ``s`` is exactly the order that ``format_trajectories`` consumes from
+        # ``base_dataset.schema_df``.
+        per_sample_batches: dict[int, list[torch.Tensor]] = {s: [] for s in range(n_samples)}
+        for pred in predictions:
+            tokens = pred["tokens"]
+            sample_idxs = pred["sample_idxs"]
+            for s in range(n_samples):
+                mask = sample_idxs == s
+                if bool(mask.any()):
+                    per_sample_batches[s].append(tokens[mask])
+
+        for sample, out_fp in sample_paths.items():
             if out_fp.is_file() and not cfg.do_overwrite:
                 logger.info(f"Skipping {out_fp} as it already exists.")
                 continue
-            else:
-                out_fp.parent.mkdir(parents=True, exist_ok=True)
-
-            seed = hash_based_seed(cfg.get("seed", None), split, sample)
-
-            logger.info(f"Generating trajectories for {split} sample {sample} to {out_fp} with seed {seed}.")
-
-            seed_everything(seed, workers=True)
-            predictions = trainer.predict(model=M, dataloaders=dataloader)
-            predictions_df = format_trajectories(dataloader.dataset, predictions)
-
+            logger.info(f"Writing {sample} sample for split {split} to {out_fp}.")
+            predictions_df = format_trajectories(base_dataset, per_sample_batches[sample])
             pa_table = GeneratedTrajectorySchema.align(predictions_df.to_arrow())
             pq.write_table(pa_table, out_fp)
 
--- a/src/MEDS_EIC_AR/__main__.py
+++ b/src/MEDS_EIC_AR/__main__.py
@@ -188,7 +188,7 @@ def generate_trajectories(cfg: DictConfig):
     if cfg.get("seed", None):
         seed_everything(cfg.get("seed", 1), workers=True)
 
-    n_samples = inference.N_trajectories_per_task_sample
+    n_trajectories = inference.N_trajectories_per_task_sample
 
     for split in inference.generate_for_splits:
         if split == train_split:
@@ -200,28 +200,33 @@ def generate_trajectories(cfg: DictConfig):
         else:
             raise ValueError(f"Unknown split {split}.")
 
-        # Skip work for samples whose output parquet already exists. If every requested sample is
-        # already on disk and ``do_overwrite`` is false, skip the predict pass entirely; otherwise
-        # we still run a single pass over the full ``N``-expanded dataset and just don't write the
-        # parquets that already exist. (Partial-skip support is a minor wrinkle — it keeps existing
-        # checkpointed runs idempotent without making us special-case mid-run resumption.)
-        sample_paths = {
-            sample: Path(cfg.output_dir) / split / f"{sample}.parquet" for sample in range(n_samples)
+        # Skip work for trajectories whose output parquet already exists. If every requested
+        # trajectory is already on disk and ``do_overwrite`` is false, skip the predict pass
+        # entirely; otherwise we still run a single pass over the full ``N``-expanded dataset and
+        # just don't write the parquets that already exist. (Partial-skip support is a minor
+        # wrinkle — it keeps existing checkpointed runs idempotent without making us special-case
+        # mid-run resumption.)
+        trajectory_paths = {
+            trajectory_idx: Path(cfg.output_dir) / split / f"{trajectory_idx}.parquet"
+            for trajectory_idx in range(n_trajectories)
         }
-        for sample_fp in sample_paths.values():
-            sample_fp.parent.mkdir(parents=True, exist_ok=True)
-        if not cfg.do_overwrite and all(p.is_file() for p in sample_paths.values()):
-            logger.info(f"Skipping all {n_samples} samples for split {split}: every parquet exists.")
+        for trajectory_fp in trajectory_paths.values():
+            trajectory_fp.parent.mkdir(parents=True, exist_ok=True)
+        if not cfg.do_overwrite and all(p.is_file() for p in trajectory_paths.values()):
+            logger.info(
+                f"Skipping all {n_trajectories} trajectories for split {split}: every parquet exists."
+            )
             continue
 
-        # Expand the base dataset so each subject contributes ``n_samples`` consecutive rows. See
-        # issue #89 for the motivation: one predict pass instead of ``N``, tighter padding, and
-        # prefix-cache reuse on backends that have one (#88, #97). The ordering invariant — subject
-        # changes slow, sample changes fast — means rows for sample ``s`` extracted from each batch
-        # in order land in subject-index order overall, which is what ``format_trajectories`` needs
-        # so its sequential ``schema_df.slice(...)`` lines up with the right subject metadata.
+        # Expand the base dataset so each subject contributes ``n_trajectories`` consecutive rows.
+        # See issue #89 for the motivation: one predict pass instead of ``N``, tighter padding,
+        # and prefix-cache reuse on backends that have one (#88, #97). The ordering invariant —
+        # subject changes slow, trajectory_idx changes fast — means rows for trajectory ``t``
+        # extracted from each batch in order land in subject-index order overall, which is what
+        # ``format_trajectories`` needs so its sequential ``schema_df.slice(...)`` lines up with
+        # the right subject metadata.
         base_dataset = base_loader.dataset
-        expanded_dataset = RepeatedPredictionDataset(base_dataset, n_samples=n_samples)
+        expanded_dataset = RepeatedPredictionDataset(base_dataset, n_trajectories=n_trajectories)
         expanded_loader = DataLoader(
             expanded_dataset,
             batch_size=base_loader.batch_size,
@@ -233,37 +238,41 @@ def generate_trajectories(cfg: DictConfig):
 
         seed = hash_based_seed(cfg.get("seed", None), split)
         logger.info(
-            f"Generating {n_samples} trajectories for each of {len(base_dataset)} subjects in split "
-            f"{split} (one interleaved predict pass over {len(expanded_dataset)} expanded rows, "
-            f"seed={seed})."
+            f"Generating {n_trajectories} trajectories for each of {len(base_dataset)} subjects "
+            f"in split {split} (one interleaved predict pass over {len(expanded_dataset)} "
+            f"expanded rows, seed={seed})."
         )
         seed_everything(seed, workers=True)
         predictions = trainer.predict(model=M, dataloaders=expanded_loader)
 
-        # Demux the flat predictions into per-sample, per-batch token lists. Within each batch the
-        # rows for sample ``s`` are in subject-index order (because the expanded dataset was built
-        # with subject-changes-slow ordering and ``shuffle=False``), and across batches the
-        # subject-index ranges are non-overlapping and increasing — so the concatenation per sample
-        # ``s`` is exactly the order that ``format_trajectories`` consumes from
+        # Demux the flat predictions into per-trajectory, per-batch token lists. Within each batch
+        # the rows for trajectory ``t`` are in subject-index order (because the expanded dataset
+        # was built with subject-changes-slow ordering and ``shuffle=False``), and across batches
+        # the subject-index ranges are non-overlapping and increasing — so the concatenation per
+        # trajectory ``t`` is exactly the order that ``format_trajectories`` consumes from
         # ``base_dataset.schema_df``.
-        per_sample_batches: dict[int, list[torch.Tensor]] = {s: [] for s in range(n_samples)}
+        #
+        # ``trajectory_idxs`` is a [B] long tensor recording, for each batch row, which of the N
+        # trajectories-per-subject that row corresponds to (0..N-1). It's distinct from
+        # ``subject_idxs`` which records the base-dataset index the row came from.
+        per_trajectory_batches: dict[int, list[torch.Tensor]] = {t: [] for t in range(n_trajectories)}
         for pred in predictions:
             tokens = pred["tokens"]
-            sample_idxs = pred["sample_idxs"]
-            # Iterate over the samples actually present in this batch rather than always doing N
-            # boolean compares. For batches that cover every sample (the common case when
-            # batch_size >= N), this is the same work; for tail batches or small batch sizes, it
-            # scales with the number of distinct sample_idxs in the batch instead.
-            for s in sample_idxs.unique().tolist():
-                mask = sample_idxs == s
-                per_sample_batches[s].append(tokens[mask])
+            trajectory_idxs = pred["trajectory_idxs"]
+            # Iterate over the trajectories actually present in this batch rather than always
+            # doing N boolean compares. For batches that cover every trajectory (the common case
+            # when batch_size >= N), this is the same work; for tail batches or small batch sizes,
+            # it scales with the number of distinct trajectory_idxs in the batch instead.
+            for t in trajectory_idxs.unique().tolist():
+                mask = trajectory_idxs == t
+                per_trajectory_batches[t].append(tokens[mask])
 
-        for sample, out_fp in sample_paths.items():
+        for trajectory_idx, out_fp in trajectory_paths.items():
             if out_fp.is_file() and not cfg.do_overwrite:
                 logger.info(f"Skipping {out_fp} as it already exists.")
                 continue
-            logger.info(f"Writing {sample} sample for split {split} to {out_fp}.")
-            predictions_df = format_trajectories(base_dataset, per_sample_batches[sample])
+            logger.info(f"Writing trajectory {trajectory_idx} for split {split} to {out_fp}.")
+            predictions_df = format_trajectories(base_dataset, per_trajectory_batches[trajectory_idx])
             pa_table = GeneratedTrajectorySchema.align(predictions_df.to_arrow())
             pq.write_table(pa_table, out_fp)
 
--- a/src/MEDS_EIC_AR/__main__.py
+++ b/src/MEDS_EIC_AR/__main__.py
@@ -250,10 +250,13 @@ def generate_trajectories(cfg: DictConfig):
         for pred in predictions:
             tokens = pred["tokens"]
             sample_idxs = pred["sample_idxs"]
-            for s in range(n_samples):
+            # Iterate over the samples actually present in this batch rather than always doing N
+            # boolean compares. For batches that cover every sample (the common case when
+            # batch_size >= N), this is the same work; for tail batches or small batch sizes, it
+            # scales with the number of distinct sample_idxs in the batch instead.
+            for s in sample_idxs.unique().tolist():
                 mask = sample_idxs == s
-                if bool(mask.any()):
-                    per_sample_batches[s].append(tokens[mask])
+                per_sample_batches[s].append(tokens[mask])
 
         for sample, out_fp in sample_paths.items():
             if out_fp.is_file() and not cfg.do_overwrite:
--- a/src/MEDS_EIC_AR/generation/repeated_dataset.py
+++ b/src/MEDS_EIC_AR/generation/repeated_dataset.py
@@ -1,0 +1,146 @@
+"""Dataset wrapper + collate helper that expands one item into ``n_samples`` interleaved rows.
+
+This is the data-layer side of issue #89: rather than running ``trainer.predict`` once per
+``sample`` index and re-prefilling every subject's input ``N`` times, we expand the base dataset to
+``len(base) * N`` items where each base item contributes ``N`` *consecutive* rows. Same-subject
+rows then end up in adjacent batch positions, which gives us:
+
+1. Tighter padding (rows in the same batch are more likely to share length).
+2. Prefix-cache reuse on backends that have one (vLLM/SGLang, see #88 / #97).
+3. One ``trainer.predict`` pass instead of ``N`` (saves dataloader/worker spawn + Lightning init).
+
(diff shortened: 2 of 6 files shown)


src/MEDS_EIC_AR/training/module.py
```diff
@@ -455,5 +455,18 @@ def predict_step(self, batch: MEDSTorchBatch):
...
+
+        If the batch was produced by :func:`MEDS_EIC_AR.generation.collate_with_meta` (the path used
+        by ``MEICAR_generate_trajectories`` to interleave ``N`` samples per subject into a single
+        predict pass — see issue #89), the per-row ``(subject_idxs, sample_idxs)`` metadata is
+        forwarded along with the generated tokens so downstream code can demux the flat batch into
+        per-sample parquet files. Plain MEDSTorchBatches without metadata return just the tokens,
+        preserving the legacy contract.
         """
-        return self.model.generate(batch, **self.generation_kwargs)
+        from MEDS_EIC_AR.generation.repeated_dataset import META_ATTR
```
mmcdermott: why not have this be a top level import?

src/MEDS_EIC_AR/generation/repeated_dataset.py
```diff
@@ -0,0 +1,146 @@
...
+    items, subject_idxs, sample_idxs = zip(*raw_items, strict=True)
+    batch = base_collate(list(items))
+    meta = (
+        torch.as_tensor(subject_idxs, dtype=torch.long),
+        torch.as_tensor(sample_idxs, dtype=torch.long),
+    )
+    # ``MEDSTorchBatch`` is a dataclass without a __dict__ in older versions; use replace + a sidecar
+    # set on the returned instance. ``object.__setattr__`` works around frozen=True if applicable.
+    object.__setattr__(batch, META_ATTR, meta)
+    return batch
```
mmcdermott: why not just return two things -- the batch, unaltered, and a list of the meta information? As a tuple? Then you don't need to mess with the batch and won't be brittle to upstream changes, the meta information won't go on the GPU, but the system using the dataset can still preserve the information and use it as needed?

src/MEDS_EIC_AR/utils.py
```diff
@@ -25,7 +25,7 @@ def is_mlflow_logger(logger: Logger) -> bool:
         return False
 
 
-def hash_based_seed(seed: int | None, split: str, sample: int) -> int:
+def hash_based_seed(seed: int | None, split: str, sample: int = 0) -> int:
```
mmcdermott: Is the capability where sample was passed in still needed? Or was that just used for generation before?

src/MEDS_EIC_AR/generation/repeated_dataset.py
```diff
@@ -0,0 +1,146 @@
...
+
+    ``RepeatedPredictionDataset.__getitem__`` returns ``(item, subject_idx, sample_idx)`` tuples,
+    so the dataloader hands ``raw_items`` to this collate as a list of those tuples. We unzip the
+    metadata, run the base collate over the items, and stash a ``(subject_idxs, sample_idxs)`` pair
+    on the resulting batch under ``META_ATTR`` for the predict step to read.
+
+    ``MEDSTorchBatch`` is a frozen dataclass, so we use ``dataclasses.replace`` to attach the
+    metadata via a sidecar dict on a fresh instance — no in-place mutation, no patching of the
+    original class.
+
```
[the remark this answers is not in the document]
mmcdermott: Fixed automatically by the tuple refactor in 9095403 — the whole sidecar-attach code path is gone. The new `collate_with_meta` returns `(batch, subject_idxs, sample_idxs)` as a plain three-tuple, so there's no ``object.__setattr__`` trick and no `dataclasses.replace` fiction left in the docstring. See the top-level response on this PR for the full change set.

src/MEDS_EIC_AR/__main__.py
```diff
@@ -180,34 +188,79 @@ def generate_trajectories(cfg: DictConfig):
...
+        # ``s`` is exactly the order that ``format_trajectories`` consumes from
+        # ``base_dataset.schema_df``.
+        per_sample_batches: dict[int, list[torch.Tensor]] = {s: [] for s in range(n_samples)}
+        for pred in predictions:
+            tokens = pred["tokens"]
+            sample_idxs = pred["sample_idxs"]
+            for s in range(n_samples):
+                mask = sample_idxs == s
+                if bool(mask.any()):
+                    per_sample_batches[s].append(tokens[mask])
```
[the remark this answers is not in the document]
mmcdermott: Good call, done in 9095403:

```python
for s in sample_idxs.unique().tolist():
    mask = sample_idxs == s
    per_sample_batches[s].append(tokens[mask])
```

For batches that cover every sample (the common case when `batch_size >= N`), the work is identical to the old `for s in range(n_samples)` loop — the bottleneck is the boolean compare, not the iteration. For tail batches or small batch sizes it now scales with the number of distinct `sample_idxs` actually present instead of always N, which matters most for the production research configs where `N_trajectories_per_task_sample` is 20+. Dropped the separate `bool(mask.any())` guard since `unique()` never returns values absent from the tensor, so each iteration is guaranteed to write something.

src/MEDS_EIC_AR/training/module.py
mmcdermott: Subsumed by the tuple refactor in 9095403 — the `META_ATTR` import was the reason for the local-scoped import (the module-level import would have worked fine, I just hadn't moved it up yet). With the sidecar approach gone, there's no `META_ATTR` symbol to import at all. `predict_step` now detects the tuple shape inline:

```python
def predict_step(self, batch):
    if isinstance(batch, tuple) and len(batch) == 3:
        batch, subject_idxs, sample_idxs = batch
        tokens = self.model.generate(batch, **self.generation_kwargs)
        return {"tokens": tokens, "subject_idxs": subject_idxs, "sample_idxs": sample_idxs}
    return self.model.generate(batch, **self.generation_kwargs)
```

No imports needed beyond what the rest of the file already pulls in. The legacy bare-MEDSTorchBatch path is preserved.

src/MEDS_EIC_AR/generation/repeated_dataset.py
mmcdermott: You're right, this is cleaner and I switched to it in 9095403. The prior sidecar-attach design (via `object.__setattr__` on the MEDSTorchBatch) was working but brittle to exactly the scenarios you flagged — if `meds_torchdata` ever made MEDSTorchBatch strictly frozen (or switched to slots) the sidecar attach would break, and the metadata was always a second-class citizen of the batch.

The tuple version:

```python
def collate_with_meta(raw_items, base_collate) -> tuple[MEDSTorchBatch, torch.Tensor, torch.Tensor]:
    items, subject_idxs, sample_idxs = zip(*raw_items, strict=True)
    batch = base_collate(list(items))
    return (
        batch,
        torch.as_tensor(subject_idxs, dtype=torch.long),
        torch.as_tensor(sample_idxs, dtype=torch.long),
    )
```

Benefits:
- **Base batch untouched.** No `object.__setattr__` hack, no assumption about whether `MEDSTorchBatch` allows extra attrs, no breakage if upstream adds `frozen=True`.
- **Meta is first-class.** It's right there in the tuple type signature; callers know what they're getting without reading docstrings about sidecar attributes.
- **Simpler destructuring on the consumer side.** `predict_step` just does `batch, subject_idxs, sample_idxs = yielded` when the input is a 3-tuple.

On "meta won't go on the GPU" — technically Lightning's `move_data_to_device` will recursively move the tensors in the tuple, so the meta tensors *do* get moved to the accelerator. I could keep them as plain Python lists to prevent that, but the tensors are tiny (`[B]` int64, maybe 128 bytes for typical batch sizes) and keeping them as tensors lets `predict_step` and the demux loop use `.unique()` / boolean masks directly. If the GPU-move cost ever shows up in a profile, we can swap to Python lists without changing the call sites. Flagging as a soft follow-up rather than blocking this PR on it.

Dropped `META_ATTR`, `extract_meta`, and the corresponding error-path unit test since they're no longer needed. The wire-up test (`test_collate_with_meta_round_trip_through_dataloader`) still covers the same end-to-end flow, just destructures a tuple instead of calling `extract_meta`.

src/MEDS_EIC_AR/utils.py
mmcdermott: Right — it's no longer needed. Post-#89, the only caller is `hash_based_seed(cfg.get("seed", None), split)` with no sample arg at all, because the new design is one seed per split (one interleaved predict pass instead of N loops). Dropped the parameter in 9095403:

```python
def hash_based_seed(seed: int | None, split: str) -> int:
    hash_str = f"{seed}_{split}"
    return int(sha256(hash_str.encode()).hexdigest(), 16) % (2**32 - 1)
```

Simpler API, simpler hash input. Doctest outputs regenerated for the new hash strings. Safe to remove because `hash_based_seed` isn't exported from `MEDS_EIC_AR/__init__.py` — the only callers are in-package, and `__main__.py` is already updated.

src/MEDS_EIC_AR/training/module.py
```diff
@@ -447,13 +447,25 @@ def configure_optimizers(self):
 
         return {"optimizer": optimizer, "lr_scheduler": LR_config}
 
-    def predict_step(self, batch: MEDSTorchBatch):
+    def predict_step(self, batch):
```
mmcdermott: still include a type hint, just reflect what it is.  You can even define a type alias in repeated_dataset for the output type of that function.

src/MEDS_EIC_AR/training/module.py
```diff
@@ -447,13 +447,25 @@ def configure_optimizers(self):
...
+
+        If the dataloader was built with :func:`MEDS_EIC_AR.generation.collate_with_meta` (the path
+        used by ``MEICAR_generate_trajectories`` to interleave ``N`` samples per subject into a
+        single predict pass — see issue #89), it yields a three-tuple
+        ``(batch, subject_idxs, sample_idxs)`` instead of a bare ``MEDSTorchBatch``. We detect that
+        shape and forward the metadata alongside the generated tokens so downstream code can demux
+        the flat batch into per-sample parquet files. Plain MEDSTorchBatches without metadata return
+        just the tokens, preserving the legacy contract.
         """
+        if isinstance(batch, tuple) and len(batch) == 3:
```
mmcdermott: actually, do we even need to support two pathways here? If someone wants to just sample trajectories a single time per patient, they can just set n_samples to 1, right? I think we can just make predict only accept the new kind of generation batches and never a batch alone without issue.

src/MEDS_EIC_AR/__main__.py
```diff
@@ -180,34 +188,82 @@ def generate_trajectories(cfg: DictConfig):
...
+        # Demux the flat predictions into per-sample, per-batch token lists. Within each batch the
+        # rows for sample ``s`` are in subject-index order (because the expanded dataset was built
+        # with subject-changes-slow ordering and ``shuffle=False``), and across batches the
+        # subject-index ranges are non-overlapping and increasing — so the concatenation per sample
+        # ``s`` is exactly the order that ``format_trajectories`` consumes from
+        # ``base_dataset.schema_df``.
+        per_sample_batches: dict[int, list[torch.Tensor]] = {s: [] for s in range(n_samples)}
+        for pred in predictions:
+            tokens = pred["tokens"]
+            sample_idxs = pred["sample_idxs"]
```
mmcdermott: what is sample_idxs? That variable name isn't especially clear.

mmcdermott (commented):
Do we have an end to end script based test that calls the generation script and verifies that file writing and such all work as expected? If we had one originally, it should still apply here, provided n_samples is > 1, I just want to double check that we do have that.

src/MEDS_EIC_AR/training/module.py
mmcdermott: Good point — added a proper type hint in e667587. Defined a `PredictBatch` alias at module level in `generation.repeated_dataset` so the collate return type and the `predict_step` input type share a single source of truth:

```python
# in src/MEDS_EIC_AR/generation/repeated_dataset.py
PredictBatch = tuple["MEDSTorchBatch", torch.Tensor, torch.Tensor]

# in src/MEDS_EIC_AR/training/module.py
from ..generation.repeated_dataset import PredictBatch

def predict_step(self, batch: PredictBatch) -> dict[str, torch.Tensor]:
    batch, subject_idxs, trajectory_idxs = batch
    tokens = self.model.generate(batch, **self.generation_kwargs)
    return {"tokens": tokens, "subject_idxs": subject_idxs, "trajectory_idxs": trajectory_idxs}
```

Combined with the next comment (drop the bare-batch branch), the signature is now unambiguous about what it accepts and returns. Anyone looking at the type hint alone can see the full contract.

src/MEDS_EIC_AR/training/module.py
mmcdermott: Agreed, simplified in e667587. Dropped the `isinstance(batch, tuple) and len(batch) == 3` branch and the bare-MEDSTorchBatch fallback:

```python
def predict_step(self, batch: PredictBatch) -> dict[str, torch.Tensor]:
    batch, subject_idxs, trajectory_idxs = batch
    tokens = self.model.generate(batch, **self.generation_kwargs)
    return {"tokens": tokens, "subject_idxs": subject_idxs, "trajectory_idxs": trajectory_idxs}
```

The only `trainer.predict` caller in the codebase (`MEICAR_generate_trajectories`) always feeds it the expanded dataset via `collate_with_meta`, so the bare-batch path was dead. Someone who wants "just one trajectory per subject" sets `n_trajectories=1` on `RepeatedPredictionDataset` and still gets a 3-tuple shape (with a trivial trajectory axis). The simpler signature is a strict improvement — less surface area, no type-detection branching, and the type checker now sees the contract rather than having to guess.

Grepped the whole tree before making the change: no other `trainer.predict` or `predict_step` callers in `src/`, `tests/`, or `conftest.py`. If any downstream repo relies on passing bare MEDSTorchBatches to `predict_step`, the update is to wrap through `collate_with_meta` — trivial, and in exchange they get the per-trajectory metadata they probably wanted anyway.

src/MEDS_EIC_AR/__main__.py
mmcdermott: Fair — "sample" is overloaded with the generic ML sense of "a batch row / datapoint" so it reads ambiguously here. Renamed throughout in e667587 to `trajectory_idx` / `trajectory_idxs` / `n_trajectories`. The Hydra config key is `inference.N_trajectories_per_task_sample`, which parses as "N trajectories per task sample" (where "task sample" = the input prompt) — so "trajectory" is the outer noun and unambiguous: it's which of the N trajectories-per-subject this row corresponds to.

Full rename covers:
- `RepeatedPredictionDataset(n_samples=N)` → `RepeatedPredictionDataset(n_trajectories=N)`
- Dataset `__getitem__` yields `(item, subject_idx, trajectory_idx)` (was `sample_idx`)
- `collate_with_meta` returns `(batch, subject_idxs, trajectory_idxs)` (was `sample_idxs`)
- `predict_step` returns `{"tokens", "subject_idxs", "trajectory_idxs"}` (was `sample_idxs`)
- CLI loop variables: `n_samples` → `n_trajectories`, `per_sample_batches` → `per_trajectory_batches`, `sample_paths` → `trajectory_paths`, `sample` → `trajectory_idx`
- Test variables mirror the rename

**Unchanged on purpose:** the output parquet filenames (`0.parquet`, `1.parquet`, ...) — the file numbering is now by `trajectory_idx` but the on-disk layout is identical, so external consumers reading these files don't see any change. And the Hydra config key `N_trajectories_per_task_sample` is unchanged since it's a user-facing API.

mmcdermott:
Yes, we do — and the existing pair still applies here unchanged. Both tests drive the full `MEICAR_generate_trajectories` CLI via `subprocess.run` and inspect the resulting parquet files, and both run with **N > 1** via the demo inference config.

`tests/test_generate_trajectories.py::test_generate_trajectories_runs` (+ `test_generate_trajectories_rolling_runs`) consume the `generated_trajectories` / `generated_trajectories_rolling` session fixtures in `conftest.py`. The fixtures spawn the CLI like so:

```python
run_and_check([
    "MEICAR_generate_trajectories",
    "--config-name=_demo_generate_trajectories",
    f"output_dir={output_dir!s}",
    f"model_initialization_dir={model_initialization_dir!s}",
    ...
    "datamodule.batch_size=2",
    "trainer=demo",
])
```

The `_demo_generate_trajectories` config extends `inference: demo`, which sets:

```yaml
# src/MEDS_EIC_AR/configs/inference/demo.yaml
generate_for_splits: [train, tuning, held_out]
N_trajectories_per_task_sample: 2
```

So every test run:
1. Spawns the generation CLI as a subprocess (no in-process shortcuts).
2. Generates **N=2 trajectories per subject** across all three splits.
3. Asserts one parquet per `(split, trajectory_idx)` on disk, all non-empty.
4. In `test_generate_trajectories_runs`: additionally asserts the two trajectories per split are **distinct** (not accidentally identical) and that the two parquets cover the **same subject set** (both trajectory indices touch every subject).
5. In `test_generate_trajectories_rolling_runs`: additionally asserts the rolling path emits trajectories longer than the single-chunk cap, proving the rolling machinery engaged.

Critically, the `batch_size=2` override combined with `N_trajectories_per_task_sample=2` means each batch is **exactly one subject's two trajectories** — which is the shape this PR most changes behavior for. If the interleaved-collate / demux pipeline were miswired, we'd see:
- wrong rows in wrong parquets (a subject's trajectory 0 accidentally written to `1.parquet`), or
- duplicate rows across trajectories (both parquets showing the same sequence for the same subject), or
- missing subjects in one of the two parquets.

The existing assertions (same subject set in both parquets, distinct DataFrames, non-empty per split) catch all three of those failure modes. I verified on this branch that both tests pass after the rename + tuple-batch refactor.

There's also a new unit test `test_collate_with_meta_round_trip_through_dataloader` that covers the collate path in isolation with a fake dataset — N=4, `batch_size=5` (deliberately not a multiple of N so we exercise the cross-subject batch case), asserts subject/trajectory indices round-trip through the real DataLoader.

No new end-to-end test is strictly needed for this PR, but if you want an additional one (e.g., N>=5 to exercise batch shapes the demo config doesn't hit), happy to add it — let me know.

PR #106, opened 2026-04-16
test: end-to-end CLI grammar integration (closes #105)

### LINKED ISSUE #105: Test: extend synthetic-grammar test to full CLI pipeline (end-to-end generation integration)
## Summary

Extend the synthetic-grammar integration test (landed with #93 in `tests/test_pattern_generation.py`) to run through the full `MEICAR_pretrain` → `MEICAR_generate_trajectories` CLI pipeline instead of driving `Model.generate` in-process, and validate the output parquet files using the existing `GrammarFSM`. This gives us a true end-to-end correctness test for the generation pipeline — something we don't have today — with a natural home for stressing PR #103's N-trajectories interleaving at larger N and for covering CLI error paths we currently miss.

## Motivation

We currently have two related tests that each cover half of what we want:

1. **`test_pattern_generation.py`** — trains a tiny model on a synthetic grammar in-process and calls `Model.generate` directly. Proves **generation is content-correct** (grammar-valid continuations) but bypasses the whole CLI / Lightning / parquet-writing pipeline.
2. **`test_generate_trajectories.py::test_generate_trajectories_runs`** — drives the `MEICAR_generate_trajectories` CLI via subprocess on a tiny random-init pretrained model. Proves **the plumbing is wired** (files exist, subjects line up, trajectories are distinct) but can't validate content because the model hasn't learned anything.

Neither test alone can catch a class of bugs that lives in the interaction between the two — e.g., a CLI-layer regression that accidentally reorders rows between the interleaved predict pass and the per-trajectory parquet write, or a format_trajectories bug that shuffles subject metadata. PR #103's demux logic is exactly the kind of code that benefits from this test.

### A. Grammar-shaped MEDS dataset fixture

Build a minimal raw-MEDS dataset from the existing grammar. Each grammar sequence becomes one subject's timeline; each token becomes one MEDS event. The code-metadata parquet registers each grammar token as a distinct MEDS code, including `TIMELINE//END` for the dummy EOS.

```
subject 0: (time_0, A[0]) (time_1, A[1]) (time_2, A[2]) (time_3, A[3]) (time_4, SEP) (time_5, B[0]) ...
subject 1: (time_0, C[0]) (time_1, C[1]) ...
```

Timestamps can be arbitrary monotonic increments; the model only consumes codes.

### B. Session-scoped fixtures

- `grammar_preprocessed_dataset`: runs `MEICAR_process_data` on the raw grammar MEDS dir, producing a tensorized cohort dir.
- `grammar_pretrained_model`: runs `MEICAR_pretrain` on that cohort with a tiny model config (2 layers, hidden 32, max_seq_len 16) for a short budget. Aim: <60s on CI.
- `grammar_generated_trajectories`: runs `MEICAR_generate_trajectories` on the pretrained model. Override `inference.N_trajectories_per_task_sample` to something larger than the default (e.g., 8) and `datamodule.batch_size` to something not a multiple of N (e.g., 3) so the interleaving logic is exercised in its less-trivial shape.

### C. Tests

1. **Grammar validity of every generated trajectory.**
   - Read each output parquet.
   - For each `(subject_id, trajectory_idx)` group, recover the token sequence (invert the code → vocab-index mapping).
   - Walk it through `GrammarFSM` and assert every transition is valid.
   - This is the real integration signal.
2. **All N parquets exist and cover all subjects.**
   - Mirrors the existing CLI test but with the larger N to catch "some trajectories missing" bugs under the new interleaving path.
3. **Rolling + grammar correctness.**
   - Run the same fixture with `rolling_generation.max_new_tokens` set large enough to force cross-chunk generation, then re-validate grammar.
4. **Distinctness.**
   - Across trajectories for the same subject, assert non-identical sequences (sampling actually diversifies).
   - Across subjects within one trajectory, assert non-identical sequences (demux actually demuxes).

### D. CLI error-path coverage

Add a few targeted tests that invoke the CLI with bad configs and assert it fails with the expected `ValueError`. These cover error paths in `__main__.py` that currently aren't hit:

- `rolling_generation.max_new_tokens: -1` → validation error.
- `rolling_generation.rolling_context_size: 4` without `max_new_tokens` → validation error.
- `rolling_generation` with an unknown key → validation error.
- Dataset with no `TIMELINE//END` code + rolling requested → error from the auto-eos check.

These are quick subprocess invocations with `pytest.raises` around the check-for-nonzero-exit pattern in `conftest.run_and_check`. Should add a dozen lines or so.

## Dependencies

None. All changes are test-side. No API changes, no new dependencies.

## Where this lives

Keep the in-process `test_pattern_generation.py` as-is — it's a fast, focused unit-style test for `Model.generate` correctness. Add the new CLI-driven tests in a new file `tests/test_pattern_generation_cli.py` (or as a new section of `test_generate_trajectories.py`, TBD). The split keeps the fast grammar test free of subprocess / training-loop overhead while adding the heavier integration coverage.

## Estimated scope

- ~150–250 lines of test code + fixture code.
- One new file for the grammar-MEDS builder, reusing `GrammarFSM`, `PROGRAMS`, `sample_sequence` from the existing pattern test.
- Runtime: target <2 min added to the full test suite (training dominates).

## Non-goals

- Not replacing `test_pattern_generation.py`. The in-process direct-model test stays as the fast feedback loop.
- Not changing any production code paths. This is purely additive test coverage.

mmcdermott:
## Summary

Adds a full-stack integration test driving ``MEICAR_process_data`` → ``MEICAR_pretrain`` → ``MEICAR_generate_trajectories`` on a synthetic grammar-formatted MEDS dataset, so generated trajectories can be validated for *content*, not just plumbing. Closes #105.

- **tests/_grammar_meds.py** — helpers that round-trip grammar tokens through MEDS format (code mapping, raw-MEDS dataset builder, FSM walker that ignores auxiliary ``TIMELINE//*`` codes).
- **conftest.py** — session-scoped grammar fixtures: raw MEDS, preprocessed, pretrained (80 epochs at hidden=32 so the model actually learns), and generated trajectories (N=8, batch_size=3 non-multiple to stress the #103 demux; rolling generation enabled so the sliding-window path is exercised).
- **tests/test_pattern_generation_cli.py** — 7 tests covering file layout, cross-trajectory diversity, vocab coverage, soft grammar-signal threshold, plus 3 CLI error-path validations (``max_new_tokens=0``, ``rolling_context_size`` without budget, unknown rolling key) that the happy-path fixtures don't hit.

### Why soft grammar signal vs. strict validity?
Strict 100%-valid grammar output is already covered by ``test_pattern_generation.py`` via direct ``Model.generate`` calls with a larger training budget. Here the CLI training budget is constrained by test runtime, so we assert ≥30% valid transitions on the best trajectory — enough to catch entire classes of CLI bugs (wrong vocab indexing, token→code mapping drift) while staying robust to sampling noise.

## Test plan

- [x] ``uv run pytest tests/test_pattern_generation_cli.py`` — 7/7 pass (~2m27s)
- [x] ``uv run pytest`` — 65/65 full suite pass (~3m44s)
- [x] Pre-commit hooks pass on all three modified files


tests/test_pattern_generation_cli.py
```diff
@@ -0,0 +1,292 @@
...
+from tests._grammar_meds import (
+    GRAMMAR_CODES,
+    GrammarFSM,
+    grammar_tokens_from_output_df,
+)
+
+# Auxiliary codes that MEICAR_process_data emits for us (timestamps → time-delta tokens; timeline
+# start/end sentinels). These aren't part of the grammar but are expected in the output and must
+# not cause a grammar-check failure.
+_AUX_CODE_PREFIXES = ("TIMELINE//", "GRAMMAR//")
```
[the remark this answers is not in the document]
mmcdermott: Good catch — fixed in 6ed9d8e. Removed `"GRAMMAR//"` from `_AUX_CODE_PREFIXES` so the test now relies on exact membership in `GRAMMAR_CODES` for grammar codes. Added a comment explaining why `GRAMMAR//` is *not* listed as an aux prefix.

tests/test_pattern_generation_cli.py
```diff
@@ -0,0 +1,292 @@
...
+        for t, df in trajectories.items():
+            per_subject: dict[int, tuple[str, ...]] = {}
+            for subject_id, subject_df in df.group_by("subject_id"):
+                per_subject[subject_id[0]] = tuple(subject_df["code"].to_list())
+            per_traj_per_subject[t] = per_subject
+
+        subjects_any = list(per_traj_per_subject["0"].keys())
+        # At least one subject must have at least two distinct trajectories out of N.
+        any_subject_is_diverse = any(
+            len({per_traj_per_subject[t].get(s) for t in per_traj_per_subject}) > 1 for s in subjects_any
```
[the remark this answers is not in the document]
mmcdermott: Valid concern — fixed in 6ed9d8e. The distinctness test now asserts that all N trajectories share the same subject set before comparing, then indexes directly (no `.get(...)`), so a dropped subject raises `KeyError` instead of silently counting as `None` → distinct. The structural test above covers the subject-set equality for the happy path, but keeping the assertion inline makes this test self-contained.

tests/test_pattern_generation_cli.py
```diff
@@ -0,0 +1,292 @@
...
+            tokens_by_subject = grammar_tokens_from_output_df(df)
+            for _subject_id, tokens in tokens_by_subject.items():
+                if not tokens:
+                    continue
+                valid, total = _walk_grammar_tokens_ignoring_start(tokens)
+                fraction = valid / total
+                best_fraction = max(best_fraction, fraction)
+
+    assert best_fraction >= 0.3, (
+        f"No generated trajectory in any split had >= 30% valid grammar transitions "
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 6ed9d8e by restricting the loop to `held_out_split` (matching the docstring rather than loosening it). Stronger signal too: train-set memorization could mask a broken pipeline, but held-out subjects were never seen during pretraining so any grammar signal there is real generalization.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,212 @@
...
+    prediction_time_offset_hours: int = 5,
+) -> None:
+    """Write a raw-MEDS directory of grammar-generated subject timelines.
+
+    Each subject's timeline is one grammar sequence sampled from
+    :func:`tests.test_pattern_generation.sample_sequence`. Tokens become MEDS events with
+    timestamps at fixed 1-hour spacing. Codes follow :data:`TOKEN_TO_CODE`. The output layout
+    matches what :class:`meds_testing_helpers.dataset.MEDSDataset` writes (one parquet per
+    split, a ``code_metadata.parquet``, and a ``subject_splits.parquet``), so
+    ``MEICAR_process_data`` can consume it without modification.
```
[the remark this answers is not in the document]
mmcdermott: Confirmed `metadata/codes.parquet` is the correct MEDS-spec filename (matches `meds.code_metadata_filepath`), and the preprocessing pipeline consumes it successfully. Fixed the docstring in 6ed9d8e to reference `metadata/codes.parquet` instead of `code_metadata.parquet`.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,212 @@
...
+                )
+            # One task-labels row per subject — prediction_time sits inside the timeline so the
+            # generator has real history to condition on and real future to generate. The label
+            # value is unused by the generation path (generate_trajectories ignores it) but the
+            # column is required by MEDSTorchDataConfig.
+            split_task_rows.append(
+                {
+                    "subject_id": subject_id,
+                    "prediction_time": base_time + timedelta(hours=prediction_time_offset_hours),
+                    "boolean_value": False,
```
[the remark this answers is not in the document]
mmcdermott: Good catch — the shortest grammar sequence (`B` + `SEP` = 4 tokens, last event at hour 3) puts the fixed 5-hour offset past the last event. Fixed in 6ed9d8e: `prediction_time` is now placed per-subject at `len(tokens) // 2` hours. This guarantees at least one history event and at least one future event regardless of sampled length, and removes the now-unused `prediction_time_offset_hours` kwarg.

mmcdermott:
broader comments from another LLM
Yes. My read is: **good direction, but I’d still want a couple changes before merge**.

What’s good:

* It fills a real gap between `test_pattern_generation.py` and the existing CLI plumbing tests.
* Stressing `N_trajectories_per_task_sample=8` with `batch_size=3` is the right shape for catching demux bugs.
* The grammar→MEDS round-trip helper is a sensible way to get content-level signal through the full CLI path.
* The three config-validation subprocess tests are useful and belong here.

What I’d comment on the PR:

1. **`test_grammar_cli_model_shows_grammar_signal` is too weak to be the main correctness assertion.**
   The current criterion is basically “the *best* held-out subject/trajectory had at least 30% valid transitions.” That is a very forgiving statistic:

   * it is an outlier metric, not a distributional one;
   * it can pass on one lucky sample while most outputs are garbage;
   * short sequences make it even easier to pass accidentally.

   I would not treat that as a strong end-to-end content regression test. If this is meant to justify closing `#105`, I’d tighten it. Better options:

   * require a threshold on the **mean** or **median** validity fraction across held-out subjects/trajectories;
   * or require **at least K held-out subjects** to exceed the threshold;
   * and ideally require a **minimum number of grammar tokens** before a sample counts.

2. **It requests rolling generation, but does not really test rolling-specific content correctness.**
   The fixture sets `rolling_generation.max_new_tokens=30`, so the rolling path should be active, but the assertions here do not show:

   * that this output actually required rolling rather than fitting in one chunk, or
   * that grammar signal is preserved specifically across the rolling path.

   You already have a separate rolling plumbing test elsewhere, but this PR is supposed to be the grammar-backed end-to-end test. I’d add either:

   * a non-rolling grammar fixture plus a comparison showing the rolling one exceeds the single-chunk cap, or
   * an explicit assertion on this fixture that at least one generated sample is longer than the non-rolling maximum while still satisfying the grammar criterion.

3. **One of the issue’s error-path acceptance criteria is still missing.**
   `#105` called out the “rolling requested but dataset/model cannot provide `TIMELINE//END` / auto-eos setup fails” path. This PR covers:

   * zero `max_new_tokens`,
   * `rolling_context_size` without budget,
   * unknown rolling key,

   but not the missing-end-token / auto-eos failure branch in `__main__.py`. I’d either add that test or stop claiming the error-path coverage is complete.

Non-blocking:

* `grammar_tokens_from_output_df` is implicitly using parquet row order as generation order. That matches current implementation, but it is a fairly brittle contract. At minimum I’d document that assumption in the helper; longer-term an explicit generation-order field would be cleaner.
* `_load_trajectories_by_split` now exists in two test files and could be shared.

So my bottom line would be:

> Strong addition overall and much closer to the right end-to-end test story, but I’d still ask for (1) a stronger grammar-quality assertion, (2) a rolling-specific content check rather than only a rolling fixture, and (3) the missing no-`TIMELINE//END` error-path test before calling this done.

tests/test_pattern_generation_cli.py
```diff
@@ -0,0 +1,306 @@
...
+    for a given subject. This is a weak but important sanity check that sampling actually diversifies.
+    """
+    by_split = _load_trajectories_by_split(grammar_generated_trajectories)
+
+    for split, trajectories in by_split.items():
+        # Build per-(trajectory_idx, subject_id) code tuple and compare across trajectories.
+        per_traj_per_subject: dict[str, dict[int, tuple[str, ...]]] = {}
+        for t, df in trajectories.items():
+            per_subject: dict[int, tuple[str, ...]] = {}
+            for subject_id, subject_df in df.group_by("subject_id"):
```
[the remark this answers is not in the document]
mmcdermott: Fixed in f57d9a7. Sort by (subject_id, time) first, then `group_by(..., maintain_order=True)`. The comment above the change documents why: without this, a polars reorder could make different orderings look like different sequences and spuriously pass the distinctness test.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,215 @@
...
+    :func:`tests.test_pattern_generation.sample_sequence`. Tokens become MEDS events with
+    timestamps at fixed 1-hour spacing. Codes follow :data:`TOKEN_TO_CODE`. The output layout
+    matches the MEDS spec (``meds.code_metadata_filepath`` is ``metadata/codes.parquet``): one
+    parquet per split, a ``metadata/codes.parquet``, and a ``metadata/subject_splits.parquet``,
+    so ``MEICAR_process_data`` can consume it without modification.
+
+    Args:
+        data_dir: Directory to write the MEDS files into. Created if it doesn't exist.
+        n_train, n_tuning, n_held_out: Number of subjects per split.
+        seed: RNG seed for grammar sampling + subject-id assignment.
```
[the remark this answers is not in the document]
mmcdermott: Fixed in f57d9a7. Docstring now clarifies that `seed` controls per-subject grammar sampling only; subject ids are assigned deterministically from `range(1, 10_000)` and do not depend on `seed`.

conftest.py
```diff
@@ -249,6 +249,114 @@ def generated_trajectories_rolling(
...
+# ---------------------------------------------------------------------------
+# Grammar-dataset fixtures (issue #105) — pretrain + generate on a synthetic
+# grammar dataset via the full CLI pipeline so we can validate generation
+# correctness end-to-end, not just plumbing.
+# ---------------------------------------------------------------------------
+
+
+# Training-hyperparameter overrides we pass to the grammar pretrain CLI. Centralized here so all
+# grammar fixtures agree on them — the CLI test file inlines references to these for its own
+# assertions (e.g. "generation output should not be wider than this max_seq_len"). See issue #105.
```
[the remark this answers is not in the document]
mmcdermott: Fixed in f57d9a7 by actually making the tests reuse the constants (your second suggestion). Moved the `_GRAMMAR_*` constants from `conftest.py` into `tests/_grammar_meds` as public names (`GRAMMAR_MAX_SEQ_LEN`, `GRAMMAR_N_TRAJECTORIES`, etc.) and imported them in both `conftest.py` and `tests/test_pattern_generation_cli.py`. The CLI test file now genuinely references them (e.g. the N-parquets-per-split assertion uses `GRAMMAR_N_TRAJECTORIES`, the rolling assertion uses `GRAMMAR_MAX_SEQ_LEN` and `GRAMMAR_ROLLING_MAX_NEW_TOKENS`), so the shared-constants invariant is enforced.

mmcdermott:
## Response to the LLM review comment above

Thank you for the detailed review — all three bullets land, and I've addressed them (commit f57d9a7). A combined response:

**(1) `test_grammar_cli_model_shows_grammar_signal` too weak.** Agreed. The root cause turned out to be two bugs, not just a too-weak metric:

- The **metric**: I was asserting `best >= 30%`, which is an outlier-dominated statistic that a single lucky short sample can satisfy. Your suggestion to use a distributional statistic (mean / passing-rate / K-of-N subjects) was the right call.
- The **FSM walker**: starting the FSM at state ``BETWEEN`` on every generated sample was plain wrong. Generated tokens continue from mid-program (wherever the prompt's timeline left the FSM), so the walker would hit ``BETWEEN -> A[1]`` as invalid on the first token and record 0% validity. I now walk only the suffix after the first generated ``SEP``, which unambiguously returns the FSM to ``BETWEEN``.

With both fixed, the held-out passing-rate distribution shows ~80% of (subject, trajectory) samples clearing a 50% per-sample-validity threshold — unambiguous signal. The test now asserts ``>= 50% of samples pass >= 50% validity``, which is a majority-requires-majority criterion that cannot pass on a single lucky sample.

**(2) Rolling-specific content check missing.** Agreed. Added ``test_grammar_cli_rolling_actually_rolls_and_preserves_signal``, which:

- Confirms at least one held-out sample's total length *exceeds* ``GRAMMAR_MAX_SEQ_LEN``, which proves the sliding-window loop actually appended tokens past the single-chunk ceiling (rather than the fixture merely configuring rolling and then every sample fitting in one chunk via early EOS).
- Applies the same passing-rate threshold to just the samples that rolled. Regresses loudly if sliding-window boundary handling corrupts generation.

Together these give a rolling-specific content assertion rather than a rolling-specific plumbing assertion (which already exists in ``test_generate_trajectories.py``).

**(3) Missing ``TIMELINE//END`` / auto-eos error path.** **Pushing back, partially.** The path you're referring to is ``__main__.py:167-179`` — rolling requested but ``eos_token_id`` couldn't be auto-populated, or ``eos == PAD_INDEX``. Both paths require a dataset whose preprocessed vocab lacks ``TIMELINE//END``, which ``MEICAR_process_data`` always adds. Triggering it requires one of:
   - A handwritten preprocessed tensorized cohort that skips that step (brittle to the preprocessing contract changing)
   - A custom checkpoint whose stored eos config has been post-hoc set to ``PAD_INDEX`` (also brittle)

I'd rather leave this uncovered and open a follow-up issue for a proper "malformed cohort" test harness than ship a contrived one. The rest of issue #105's error-path acceptance criteria are covered; this one I'd like to scope out explicitly rather than leave as a silent gap. Will file a follow-up if you want me to.

**Non-blocking:**

- Parquet-row-order contract in ``grammar_tokens_from_output_df``: now documented with an explicit note that a future change which reorders rows on write would break the test silently. A cleaner long-term fix would be an explicit ``generation_order`` column on the output parquet; tracking as future work.
- ``_load_trajectories_by_split`` exists only in ``test_pattern_generation_cli.py``, not in ``test_generate_trajectories.py``. Checked — the comment was off. Not duplicated, no action needed.

## On ``test_pattern_generation.py`` strength vs. this test

Separate note the author of this PR raised out-of-band: "Can this test be as strong as the in-process one, which fully agrees with the grammar?" The honest answer is **not easily at this budget**, for three compounding reasons:

1. **FSM starting state is ambiguous** — in-process tests generate from scratch (``BETWEEN``); the CLI test generates from a prompt that can end mid-program.
2. **Sampling is on** — CLI matches production (``do_sample=True``), so even a converged model will occasionally pick a wrong token. In-process tests can use greedy.
3. **Training budget is CLI-bounded** — even at the bumped fixture size (hidden=128, 400 epochs), the model isn't as converged as the in-process test's longer schedule.

Each is independently fixable (take max over all plausible starting states; add a greedy companion fixture; bump budget further). Happy to do any combination if a stricter correctness guarantee is wanted — the runtime budget is the main knob. For this PR I've landed the "unambiguous distributional signal" version (~5 min fixture runtime) and can iterate if we want to push toward near-strict validity.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+    data_dir = Path(data_dir)
+    (data_dir / "data" / "train").mkdir(parents=True, exist_ok=True)
+    (data_dir / "data" / "tuning").mkdir(parents=True, exist_ok=True)
+    (data_dir / "data" / "held_out").mkdir(parents=True, exist_ok=True)
+    (data_dir / "metadata").mkdir(parents=True, exist_ok=True)
+
+    rng = random.Random(seed)
+    base_time = datetime(2020, 1, 1, 0, 0, 0, tzinfo=UTC)
+    subject_counter = iter(range(1, 10_000))
```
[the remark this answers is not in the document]
mmcdermott: Good catch. Fixed in 761b019 by declaring the polars timestamp columns as `pl.Datetime("us", time_zone="UTC")` for both the events `time` and task-labels `prediction_time`. Matches the `tzinfo=UTC` Python datetimes used when building rows, so no silent tz-strip happens at DataFrame construction and the parquet metadata records UTC explicitly. I deliberated briefly on going naive end-to-end instead, but keeping tz-aware matches what modern MEDS tooling tends to write (and avoids a `# noqa: DTZ001` on the Python side).

conftest.py
```diff
@@ -249,6 +259,108 @@ def generated_trajectories_rolling(
...
+@pytest.fixture(scope="session")
+def grammar_pretrained(grammar_preprocessed: Path, tmp_path_factory: pytest.TempPathFactory) -> Path:
+    """Runs ``MEICAR_pretrain`` on the grammar preprocessed dataset.
+
+    Uses a bigger model than the default demo preset (hidden_size=32 via 2 heads x head_dim=16)
+    and a meaningful training budget (80 epochs x a few batches each) so the model actually has
+    a chance to learn the grammar patterns. Disables early-stopping for the same reason — the
+    demo patience=3 kicks in after only a handful of validation checks against a noisy baseline.
+
+    Expected runtime: roughly a minute on a laptop CPU. This is the bulk of the grammar
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 761b019. The `grammar_pretrained` docstring was stale — it still quoted the `hidden_size=32 via 2 heads x head_dim=16` and 80-epoch budget from the earlier iteration before the Copilot + LLM round of feedback pushed me to scale up for unambiguous signal. Docstring now defers to the `GRAMMAR_*` constants in `tests/_grammar_meds` (so we do not have to keep two copies of the numbers in sync) and explicitly names this fixture as the substrate for the strict greedy-correctness assertion, not just the distributional sampling test.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,259 @@
...
+#: Per-head dim. At small scale the model is capacity-limited on grammar generalization;
+#: hidden_size=128 (plus more training data and more epochs, below) reliably produces an
+#: unambiguous distributional held-out grammar signal that survives sampling variance.
+GRAMMAR_MODEL_HEAD_DIM: int = 32
+#: Pretrain epoch budget. Sized to produce clear, reproducible signal on held-out — this test
+#: is intentionally the expensive end-to-end content regression test for the CLI pipeline; it
+#: can be skipped locally in favor of doctests during iteration and run in CI or pre-commit
+#: sweeps when content correctness is the question.
+GRAMMAR_PRETRAIN_EPOCHS: int = 400
+#: Batch size for pretrain.
```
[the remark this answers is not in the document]
mmcdermott: Addressed in 761b019 — this and the `grammar_pretrained` docstring you flagged separately were the same drift, originating from the PR description / module docstring pre-dating the scale bump. Corrected the module docstring in `test_pattern_generation_cli.py` to match the current test layout (sampling signal + greedy strict validity + rolling content + structural + vocab + distinctness) and to drop the outdated "deliberately weaker than the in-process test" framing, which was only true of the original 80-epoch / hidden=32 fixture. The larger budget is intentional — see the rationale on `GRAMMAR_PRETRAIN_EPOCHS` which now compares forward-pass count to `test_pattern_generation.py`.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,259 @@
...
+GRAMMAR_PRETRAIN_BATCH_SIZE: int = 16
+#: N trajectories generated per task-sample in the generate CLI call.
+GRAMMAR_N_TRAJECTORIES: int = 8
+#: Batch size for the generate CLI — deliberately not a multiple of ``GRAMMAR_N_TRAJECTORIES`` so
+#: the demux from #103 is stressed in its cross-sample-group configuration.
+GRAMMAR_GENERATE_BATCH_SIZE: int = 3
+#: Rolling-generation budget for the generate CLI. Chosen strictly above the single-chunk cap
+#: ``GRAMMAR_MAX_SEQ_LEN`` so at least one sample must cross a chunk boundary; see the rolling
+#: content assertion in ``test_grammar_cli_rolling_produces_sequences_beyond_single_chunk``.
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 761b019. The test was renamed during iteration (was `test_grammar_cli_rolling_produces_sequences_beyond_single_chunk`, is now `test_grammar_cli_rolling_actually_rolls_and_preserves_signal`) and the docstring reference on `GRAMMAR_ROLLING_MAX_NEW_TOKENS` did not get updated with it. Updated now so anyone grepping for the cited assertion lands on the real test.

src/MEDS_EIC_AR/configs/inference/default.yaml
```diff
@@ -3,3 +3,8 @@ generate_for_splits:
   - held_out
 
 N_trajectories_per_task_sample: 20
+
+# Forwarded to ``Model.generate(do_sample=...)``. Default ``true`` matches production use
+# (stochastic sampling for trajectory diversity). Set ``false`` for greedy argmax decoding —
+# useful for correctness regression tests where non-determinism would obscure the signal.
+do_sample: true
```
mmcdermott, line 10: There should be a warning raised in the CLI if this is ever set to false in practice. It is an anti-pattern in real use, though I understand you need to pipe it in here for the test and am ok with that.

tests/_grammar_meds.py
mmcdermott, line 1: It seems like we should put all the grammar code in one file, rather than distributing some in one test, some here, and then having a separate test for the CLI. It also might be useful to have a `tests/grammar_tests/...` subfolder approach to isolate the files related to this from other test files.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,329 @@
...
+#: Batch size for the generate CLI — deliberately not a multiple of ``GRAMMAR_N_TRAJECTORIES`` so
+#: the demux from #103 is stressed in its cross-sample-group configuration.
+GRAMMAR_GENERATE_BATCH_SIZE: int = 3
+#: Rolling-generation budget for the generate CLI. Chosen strictly above the single-chunk cap
+#: ``GRAMMAR_MAX_SEQ_LEN`` so at least one sample must cross a chunk boundary; see the rolling
+#: content assertion in ``test_grammar_cli_rolling_actually_rolls_and_preserves_signal``.
+GRAMMAR_ROLLING_MAX_NEW_TOKENS: int = 30
+
+
+def build_grammar_meds_dataset(
```
mmcdermott: Eventually, we should think about porting this to MEDS-testing helpers, as it could be useful in many settings. (https://github.com/Medical-Event-Data-Standard/meds_testing_helpers)

It should not happen now, but you should file an upstream github issue on Meds testing helpers referencing this PR / code for the future.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,329 @@
...
+#
+# These are the knobs both the pretrain fixture (``conftest.py::grammar_pretrained``) and the
+# generate fixture (``conftest.py::grammar_generated_trajectories``) need to agree on, and the
+# assertions in ``tests/test_pattern_generation_cli.py`` also read them so we don't have two
+# copies of "N=8" that silently drift apart. Module-level constants in a test helper are the
+# least-awkward way to share these; a pytest fixture wrapping them would also work but adds
+# indirection for no gain.
+# -------------------------------------------------------------------------
+#: ``max_seq_len`` override passed to ``MEICAR_pretrain``.
+GRAMMAR_MAX_SEQ_LEN: int = 16
```
mmcdermott: Why not make many of these parameters be pytest options? I know we don't plan to change them, but it seems a bit nicer than general global constants, right? And if there is an issue with test flakiness or a change, it might be easier to adjust them to see what's going on. What do you think?

tests/_grammar_meds.py
mmcdermott: In the meantime, you might think about using meds-testing-helpers to simplify some of this code. You can make a MEDSDataset object with your generation code, then just call the "write" method to get it to disk, for example. Can you assess whether that would simplify this code here? I think it is already an existing dependency, so it wouldn't add anything to the install either.

tests/_grammar_meds.py
```diff
@@ -0,0 +1,329 @@
...
+        '{"dataset_name": "grammar_integration_test", "dataset_version": "0.1.0", '
+        '"etl_name": "tests/_grammar_meds.py", "etl_version": "0.1.0"}'
+    )
+
+    if task_labels_dir is not None:
+        # Mirror meds_testing_helpers' task_labels layout: one parquet per task-split group.
+        task_dir = Path(task_labels_dir)
+        task_dir.mkdir(parents=True, exist_ok=True)
+        for split, rows in task_labels_by_split.items():
+            task_df = pl.DataFrame(
```
mmcdermott: if this isn't in the meds-testing-helpers, you can also use direct schema objects from the meds package for some of this. It will prevent drift and reduce code burden. https://github.com/Medical-Event-Data-Standard/meds

conftest.py
```diff
@@ -249,6 +259,155 @@ def generated_trajectories_rolling(
...
+# grammar dataset via the full CLI pipeline so we can validate generation
+# correctness end-to-end, not just plumbing.
+# ---------------------------------------------------------------------------
+
+
+# Training-hyperparameter constants live in ``tests/_grammar_meds`` so pretrain and generate
+# fixtures + CLI test assertions all agree on them. See issue #105.
+
+
+@pytest.fixture(scope="session")
```
mmcdermott: Do these need to be fixtures in the conftest? If they're only used in one test file, why not just put them there? Or in a nested conftest or something in an appropriate dir. We don't ever use these in doctests, right? That's really why so many fixtures get set up like this.

src/MEDS_EIC_AR/configs/inference/default.yaml
mmcdermott: Good call. Added a warning at `__main__.py:189` that fires whenever `inference.do_sample=False` — frames it as "anti-pattern for real trajectory generation" and names the one legitimate use case (deterministic correctness tests). The grammar-CLI test fixture deliberately trips this warning; that is acceptable and the warning verbiage calls out correctness testing as the exception.

tests/_grammar_meds.py
mmcdermott: Agreed. Will move to a `tests/grammar/` subfolder with three files: `_grammar.py` (PROGRAMS / SEP / GrammarFSM / sample_sequence / build_training_batch_codes, extracted from the current `test_pattern_generation.py`), `_meds.py` (round-trip + MEDSDataset writing, refactored from `_grammar_meds.py`), and the two test files (`test_in_process.py`, `test_cli.py`). This consolidates the grammar definition in one place and scopes both test files to it. Doing it in this PR since it interacts with everything else I am touching.

tests/_grammar_meds.py
mmcdermott: Filing now — will link the issue back here when created. Marking as follow-up, not in scope for this PR.

tests/_grammar_meds.py
mmcdermott: Good idea. Going to use pytest options with the current constants as defaults — that way `--grammar-pretrain-epochs=800` works for interactive debugging without touching code, and the defaults keep the documented behavior. Exposing them as fixtures (e.g. `grammar_pretrain_epochs`) that the builder + test file consume rather than importing module-level constants.

tests/_grammar_meds.py
mmcdermott: Good suggestion — `MEDSDataset` is already a dependency via `meds_testing_helpers`. Switching to it drops the manual polars schema definitions, parquet writes for each of data/metadata/splits, and the hand-written `dataset.json` JSON literal. Using `MEDSDataset(data_shards=..., dataset_metadata=DatasetMetadataSchema(...), code_metadata=..., subject_splits=..., task_labels={...}).write(out_dir)` replaces ~60 lines with ~20. Task labels go under `task_labels/<task_name>/<shard>.parquet` per MEDSDataset convention; will update the CLI fixtures to point at the per-task subfolder.

tests/_grammar_meds.py
mmcdermott: Will use `meds.DatasetMetadataSchema` (for the metadata JSON) and `MEDSDataset.PL_LABEL_SCHEMA` (for the task-label polars schema) as part of the `MEDSDataset` refactor — those give us the canonical column names and dtypes without redefinition. The event `time` column schema stays `pl.Datetime("us", time_zone="UTC")` since MEDS spec treats it as timezone-aware.

src/MEDS_EIC_AR/configs/inference/default.yaml
mmcdermott: Done in 9517dc4. CLI warning fires whenever `inference.do_sample=false`; text frames it as an anti-pattern for real trajectory generation and names correctness testing as the only legitimate exception. The grammar-CLI greedy fixture deliberately trips this warning.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. New layout under `tests/grammar/`: `_grammar.py` (PROGRAMS, SEP, GrammarFSM, sample_sequence, build_training_batch_codes, mock_batch), `_meds.py` (MEDS round-trip via MEDSDataset), `conftest.py` (pytest options + fixtures), `test_in_process.py` (moved from the old `test_pattern_generation.py`), `test_cli.py` (moved from `test_pattern_generation_cli.py`). The three old top-level files are deleted. All 18 tests pass (5 doctests + 9 CLI + 4 in-process).

tests/_grammar_meds.py
mmcdermott: Filed upstream: https://github.com/Medical-Event-Data-Standard/meds_testing_helpers/issues/40 — references this PR. Not blocking.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `pytest_addoption` in `tests/grammar/conftest.py` registers `--grammar-pretrain-epochs`, `--grammar-n-held-out`, `--grammar-max-seq-len`, etc. (full list under the `grammar` option group). Defaults match the previous GRAMMAR_* constants. Each option has a session-scoped fixture (`grammar_pretrain_epochs` etc.) that the builder + test files consume. Example debug sweep: `pytest tests/grammar/test_cli.py --grammar-pretrain-epochs=800 --grammar-n-held-out=32`.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `build_grammar_meds_dataset` now constructs a `MEDSDataset` in memory (data_shards, dataset_metadata, code_metadata, subject_splits, task_labels) and calls `.write(out_dir)`. Drops ~60 LOC of manual parquet writes, inline `dataset.json` JSON literal, per-split directory creation, and a hand-built task-labels layout. Task labels now live at the canonical `task_labels/<task_name>/<shard>.parquet` — the fixture `grammar_raw_meds` returns both the MEDS root and the per-task subfolder so the CLI `--task_labels_dir` override picks the right path.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `DatasetMetadataSchema` from `meds` is used for the metadata object (passed into `MEDSDataset`); `MEDSDataset.PL_LABEL_SCHEMA` is used as the polars schema for task-labels rows. Canonical column-name / dtype constants from `meds` (`data_subdirectory`, `held_out_split`, etc.) are imported where needed. Event `time` column kept at `pl.Datetime("us", time_zone="UTC")` since MEDS spec treats it as timezone-aware.

src/MEDS_EIC_AR/configs/inference/default.yaml
mmcdermott: Done in 9517dc4. Warning fires at `__main__.py:190` whenever `inference.do_sample=false`; text frames it as an anti-pattern for real trajectory generation and names correctness testing as the only legitimate exception. The grammar-CLI greedy fixture deliberately trips this warning.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. New layout under `tests/grammar/`: `_grammar.py` (grammar definitions), `_meds.py` (MEDS round-trip via MEDSDataset), `conftest.py` (pytest options + fixtures), `test_in_process.py` (moved from the former `test_pattern_generation.py`), `test_cli.py` (moved from `test_pattern_generation_cli.py`). The three old top-level files are deleted. All 18 tests pass (5 doctests + 9 CLI + 4 in-process) in ~8.5 min.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `pytest_addoption` in `tests/grammar/conftest.py` registers `--grammar-pretrain-epochs`, `--grammar-n-held-out`, `--grammar-max-seq-len`, `--grammar-model-heads`, `--grammar-model-head-dim`, `--grammar-pretrain-batch-size`, `--grammar-n-trajectories`, `--grammar-generate-batch-size`, `--grammar-rolling-max-new-tokens`, `--grammar-n-train`, `--grammar-n-tuning`. Defaults match the previous GRAMMAR_* constants. Each option has a session-scoped fixture (`grammar_pretrain_epochs` etc.) that the builder + test files consume. Example debug sweep: `pytest tests/grammar/test_cli.py --grammar-pretrain-epochs=800 --grammar-n-held-out=32`.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `build_grammar_meds_dataset` now constructs a `MEDSDataset` in memory (data_shards, dataset_metadata, code_metadata, subject_splits, task_labels) and calls `.write(out_dir)`. Drops ~60 LOC of manual parquet writes, inline `dataset.json` JSON literal, per-split directory creation, and a hand-built task-labels layout. Task labels now live at the canonical `task_labels/<task_name>/<shard>.parquet` — the `grammar_raw_meds` fixture returns both the MEDS root and the per-task subfolder path so the CLI `--task_labels_dir` override picks the right path.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `DatasetMetadataSchema` from `meds` is used for the metadata object (passed into `MEDSDataset`); `MEDSDataset.PL_LABEL_SCHEMA` is used as the polars schema for task-labels rows. Canonical constants from `meds` (`data_subdirectory`, `held_out_split`, `train_split`, `tuning_split`) are imported where needed. Event `time` column kept at `pl.Datetime("us", time_zone="UTC")` since MEDS spec treats it as timezone-aware.

tests/_grammar_meds.py
mmcdermott: Done in 9517dc4. `build_grammar_meds_dataset` now constructs a `MEDSDataset` in memory (data_shards, dataset_metadata, code_metadata, subject_splits, task_labels) and calls `.write(out_dir)`. Drops the manual parquet writes, inline `dataset.json` literal, and hand-built task-labels layout. Task labels live at the canonical `task_labels/<task_name>/<shard>.parquet`; the `grammar_raw_meds` fixture returns both the MEDS root and the per-task subfolder so the CLI `--task_labels_dir` override picks the right path.

conftest.py
```diff
@@ -249,6 +249,9 @@ def generated_trajectories_rolling(
     return output_dir
 
 
+# Grammar-dataset fixtures (issue #105) live in ``tests/grammar/conftest.py``.
```
mmcdermott: You don't need this comment

tests/grammar/test_in_process.py
mmcdermott, line 1: Do we actually even need these tests anymore given the CLI tests? What do these tests cover that the CLI tests don't?

conftest.py
mmcdermott: Dropped in 386a866.

tests/grammar/test_in_process.py
mmcdermott: Fair question. My read is **keep them**, but the overlap is genuinely narrow and worth laying out.

The in-process file has four tests. Coverage vs. the CLI:

**Two FSM-only tests** (`test_grammar_fsm_rejects_hand_crafted_invalid_sequences`, `test_grammar_fsm_alt_grammar_rejects_default_grammar_sequences`): test `GrammarFSM` directly with no model, no CLI, no training. They protect against a vacuous always-accepting FSM that would make every grammar-validity assertion downstream meaningless. Zero overlap with the CLI test — the CLI test *uses* the FSM but never validates the FSM itself is non-trivial. These are cheap (~sub-second each) and unambiguously worth keeping.

**`test_trained_model_single_chunk_generation_recovers_grammar`**: uses mid-program prompts like `[A[0], A[1]]` and asserts greedy argmax emits `A[2], A[3]` next. The CLI test **cannot** cover this — after the #111 fix, `prediction_time` always lands strictly between a SEP and the next event, so the CLI test only exercises "start from BETWEEN state." If `Model.generate` regressed specifically on in-program continuation, only this in-process test would catch it. I treat that as a real coverage gap, not duplication.

**`test_trained_model_rolling_generation_preserves_grammar`**: asserts the trained models rolling output is **rejected** by `ALT_PROGRAMS`s FSM within a short prefix. The CLI test has no cross-grammar negative control — if we accidentally introduced a bug where the FSM accepts everything, the CLI assertions would all still "pass." This is the same "prove the test isnt vacuous" concern that motivates the FSM-only tests above, but applied to the trained models output distribution.

**Plus fast feedback**: ~20s for the in-process suite vs. ~8.5 min for the CLI suite. For iteration on `Model.generate` / the rolling loop specifically, the in-process suite is what Id actually use; the CLI suite is what you run in pre-commit / CI for end-to-end regression.

That said — happy to cut. Two options if you want the surface reduced:
1. Keep just the two FSM-only tests (clearest win, no model). Drop `grammar_trained_model` fixture and the two trained-model tests.
2. Keep everything (current).

(1) loses the mid-program + ALT_PROGRAMS negative control but keeps the cheap FSM invariants. Id default to (2) unless test runtime is actually biting you somewhere. Let me know and Ill do the trim if you prefer (1).

PR #116, opened 2026-04-18
feat: swap base architecture from GPT-NeoX to Llama (closes #108)

### LINKED ISSUE #108: Switch base transformer architecture from GPT-NeoX to a modern decoder (Llama-style)
## Motivation

The model is currently wrapped around HuggingFace's `GPTNeoXForCausalLM` — a 2022-era architecture. Modern decoder-only transformers (Llama, Qwen2, Mistral) are architecturally cleaner at the small scale we train at (≤ 50M params) and are better-supported across the inference-engine ecosystem. This issue is specifically about whether to switch the base architecture in-repo. It's independently valuable from the inference-backend work (#88, #97) but does conveniently unblock SGLang as a side effect.

## Recommended target: `LlamaForCausalLM`

Llama is the most vanilla modern decoder-only architecture in HF's model zoo. Relative to GPT-NeoX, at small scale:

| Component | GPT-NeoX | Llama | Notes |
|---|---|---|---|
| Normalization | LayerNorm (mean+variance, with bias) | RMSNorm (variance-only, no bias) | RMSNorm is faster and has fewer params. Ablations consistently show no quality loss. |
| MLP activation | GELU with MLP biases | SwiGLU (gated), no biases | SwiGLU is strictly stronger than GELU at matched FLOPs per PaLM / Shazeer 2020 / subsequent ablations. |
| Rotary embeddings | Partial-dim RoPE (`rotary_pct < 1`) | Full-dim RoPE | Partial-dim RoPE was a NeoX-era hack; the Llama family moved to full-dim and nothing lost. |
| Residual structure | Parallel (attn + MLP in parallel on the same residual) | Sequential | Parallel residual was a training-speed tradeoff in GPT-J/NeoX era; sequential is the modern default and trains at least as well. |
| Positional embeddings | Absolute (learned) off, RoPE on | RoPE only | Same effective stack on this axis. |
| Attention | MHA | MHA or GQA (configurable) | Llama reduces cleanly to plain MHA via `num_key_value_heads = num_attention_heads` — one-line config. |

**Expected quality impact:** modest upward. None of these are invasive changes (no architecture surgery, no retraining dataset change), but the compounding effect of RMSNorm + SwiGLU + full-dim RoPE is consistently positive across published small-scale ablations. We train on MEDS code vocab rather than text, so we're not bottlenecked on tokenizer-interaction quirks that sometimes distinguish architectures.

**Candidates considered and ruled out:**
- `Qwen2ForCausalLM`: Llama + QKV bias. Fine, but diverges from canonical Llama configs — adds a translation step if we ever benchmark against published Llama small-model recipes.
- `MistralForCausalLM`: defaults to sliding-window attention; requires `sliding_window=None` to behave like plain MHA. More fiddly for no gain.
- `GemmaForCausalLM`: non-standard embedding scaling and tied output with a scaled-output quirk; more surprises at small scale.
- `PhiForCausalLM`: partial-dim RoPE, same drawback as NeoX.

### Impacts independent of the inference backend

1. **Training configs.** The `_demo_pretrain.yaml` and any real-run configs currently instantiate `Model({gpt_kwargs...})`, which forwards to `AutoConfig.from_pretrained(\"EleutherAI/gpt-neox-20b\")` and overrides attributes. The base config swap is a one-liner in `Model.__init__` (`\"meta-llama/Llama-3.2-1B\"` or similar as the template config), plus a rename sweep: `num_attention_heads`, `num_hidden_layers`, `hidden_size`, `vocab_size`, `max_position_embeddings` transfer directly; `rotary_pct` goes away; `intermediate_size` becomes required (was implicit `4*hidden_size` in NeoX, is explicit in Llama and usually `~2.67*hidden_size * 2` for SwiGLU parity). Pretraining config review + test config update: ~1 day.
2. **Checkpoint invalidation.** Existing pretrained checkpoints become unusable since the state_dict shape changes. Weight conversion NeoX→Llama is a dead end (LayerNorm vs RMSNorm, GeLU vs SwiGLU — the SwiGLU gate projection has no counterpart in NeoX weights) — see the agent investigation summarized in #88. Anyone with a published pretrained checkpoint needs to re-run pretraining. At our scale pretraining is cheap (~minutes–hours on a single GPU for the demo configs), so this is not a blocker, but it's a one-way door: users with in-flight experiments on NeoX need to finish them before the swap lands or be prepared to rerun.
3. **Doctests.** All doctests in `src/MEDS_EIC_AR/model/model.py` that pin generated token sequences (there are several) will need to be re-seeded and regenerated under the new architecture. The Llama rotary + RMSNorm path will produce different deterministic outputs even with `torch.manual_seed(0)`. Mechanical update, but needs someone to run each doctest and paste the new expected output.
4. **Rolling-generation loop.** No change needed. `_rolling_generate` is architecture-agnostic — it operates purely on token-id tensors via `HF_model.generate`. The sliding-window positional correctness concern called out in PR #86 applies identically to NeoX and Llama (both use RoPE, both extrapolate badly past their training context).
5. **FlashAttention integration.** Unchanged — Llama's HF implementation has a FlashAttention 2 path same as NeoX, selected via `attn_implementation=\"flash_attention_2\"` (already in `Model.__init__`).
6. **Precision / dtype paths.** Unchanged — the Lightning-Module-level precision handling is architecture-agnostic.
7. **Tests.** The three existing test layers (doctests, in-process grammar test, CLI end-to-end grammar test from #106) would need to be re-run after the swap to regenerate deterministic outputs and re-confirm the grammar-signal threshold. Mechanical.

### Impacts related to the inference backend (#88, #97)

1. **Unblocks SGLang.** `GPTNeoXForCausalLM` is not in SGLang's model registry; `LlamaForCausalLM` is first-class. See the #88 investigation comment.
2. **Already first-class in vLLM.** No change to the vLLM integration plan from #97.
3. **`GenerationBackend` abstraction from PR #107 is unaffected.** The protocol is architecture-agnostic — the adapter receives an HF `PreTrainedModel` handle and a `GenerationConfig`, neither of which cares what the underlying architecture is.
4. **Tokenizer bypass (`skip_tokenizer_init`) unchanged.** Both backends support this on Llama the same way they'd support it on NeoX.

### Non-impacts worth naming explicitly

- **Data pipeline:** `MEICAR_process_data` and MEDS preprocessing are unrelated to the model architecture.
- **MEDS code vocab:** vocab size is a `Model` constructor arg — flows through to `LlamaConfig.vocab_size` exactly as it flows to `GPTNeoXConfig.vocab_size`.
- **Lightning module:** `MEICARModule` holds a `Model` instance; it doesn't introspect architecture.
- **Downstream task heads:** none currently exist; this issue doesn't change that.

## Open questions

1. **Re-pretraining cost budget.** Are we at a point where we're comfortable invalidating any existing pretrained checkpoints? If there are in-flight experiments we care about reproducing on NeoX specifically, the swap should be gated on those completing.
2. **Config surface for SwiGLU's `intermediate_size`.** Llama's `intermediate_size` is an explicit config knob; NeoX's was implicitly `4 * hidden_size`. Do we want to expose it in our `gpt_kwargs` dict or pick a sensible default (e.g., `hidden_size * 8 // 3` for SwiGLU FLOP-parity with NeoX's `4 * hidden_size` GELU MLP)?
3. **Renaming.** `Model.HF_model_config: GPTNeoXConfig` → `LlamaConfig` is a type change. We can either keep the attribute name (slight misnomer) or rename to something generic like `hf_config`. Minor bikeshed.
4. **Timing vs. #107, #88, #97.** PR #107 (backend abstraction) is independent and can merge either order. For the inference-engine work, doing the model swap *first* reduces risk (fewer moving parts in the engine PR, SGLang path unblocked).

## Recommendation / phasing

1. **Land this issue as Phase 1** — swap `Model`'s base to `LlamaForCausalLM`, update demo configs, regenerate doctests, re-run grammar tests. Independent of engine work.
2. **Then #88 / #97** — add an inference backend (vLLM recommended per #97 investigation) under the `GenerationBackend` abstraction from #107.

Happy to scope this into a single PR or split into (a) architecture swap + (b) doctest regeneration if the doctest updates get noisy.

## Labels
`enhancement`, possibly `priority:medium`

mmcdermott:
## Summary

Swaps the base HF architecture this repo wraps from `GPTNeoXForCausalLM` to `LlamaForCausalLM`. Closes #108. Unblocks #88 / #97 (SGLang / vLLM) as a side effect — both engines treat Llama as first-class.

## What changes

**Core wiring (`src/MEDS_EIC_AR/model/model.py`)**
- `HF_model_config: LlamaConfig`, `HF_model: LlamaForCausalLM`.
- Base config constructed from a vanilla `LlamaConfig()` (no network call) with caller overrides applied on top.
- After the override pass, `num_key_value_heads` is snapped to `num_attention_heads` and `head_dim` to `hidden_size // num_attention_heads` unless the caller passed them explicitly. This matches NeoX's implicit behavior (plain MHA) and keeps the same single-knob config surface callers already use. GQA opt-in is available by passing `num_key_value_heads` directly.
- Invalid-key error message now reads `Config for HF model {config.model_type} does not have attribute X` rather than hardcoding an architecture name — stays correct if we ever swap again.
- `torch_dtype` → `dtype` in the HF constructor kwargs (the old key now warns).

**Weight-decay grouping (`src/MEDS_EIC_AR/training/module.py`)**
- `_is_norm_bias_param` regex widened from `layer_?norm(\d*)\.weight` to `(layer_?)?norm(\d*)\.weight` so Llama's final `model.norm.weight` (no `layer_` prefix) stays out of the weight-decay group. **Without this, the final RMSNorm would have silently received weight decay.** Real behavioral fix, not just doctest cosmetics.

**Configs (`configs/lightning_module/model/{default,demo}.yaml`)**
- Drop `hidden_dropout` and `classifier_dropout` — neither exists on `LlamaConfig`.
- Keep `attention_dropout` (Llama has it).
- `intermediate_size` formula kept at `4 * hidden_size`. Documented in-file that strict NeoX parameter-parity under SwiGLU would need `8/3 * hidden_size`; callers chasing parity can override explicitly.

## One-way door: checkpoint invalidation

Per issue #108, existing NeoX pretrained checkpoints become unusable (state_dict shapes change; SwiGLU has no NeoX counterpart to convert from). At this repo's scale pretraining is cheap (minutes on a single GPU for the demo configs), so this is a one-way swap not gated on a migration path. Users with in-flight NeoX experiments should finish them on `main` before pulling this branch.

## Verification

**Doctests (8 in model.py, 8 in training/module.py)**: all regenerated under Llama. Loss, logits, and parameter-name outputs all shifted under RMSNorm + SwiGLU + full-dim RoPE. The `_check_parameters` doctest was retargeted from `query_key_value.bias` (Llama has no biases) to an RMSNorm weight.

**Grammar test (in-process, `tests/grammar/test_in_process.py`)**: 4/4 pass.

**Grammar test (full CLI, `tests/grammar/test_cli.py`)**: 9/9 pass, including:
- Strict 100% greedy grammar validity (every generated token a valid FSM continuation from the prompt's end-state).
- Sampling-fixture distributional signal above the 50%-of-samples-at-50%-validity threshold.
- Rolling-path content signal at the same threshold.

**Full non-CLI suite**: 63 tests pass (`uv run pytest --doctest-modules src/ tests/ -q --ignore=tests/grammar/test_cli.py`).

Agnostic-by-construction: the rolling-generation loop and everything downstream of `HF_model` are architecture-independent (they operate on token-id tensors and HF's `GenerationConfig`). The swap is contained to `Model.__init__`, config files, and doctest outputs.

## Test plan

- [x] `uv run pytest --doctest-modules src/MEDS_EIC_AR/model/model.py` — 8 pass
- [x] `uv run pytest --doctest-modules src/MEDS_EIC_AR/training/module.py` — 8 pass
- [x] `uv run pytest tests/test_generation_backend.py` — 6 pass (backend protocol still honored)
- [x] `uv run pytest tests/grammar/test_in_process.py` — 4 pass
- [x] `uv run pytest tests/grammar/test_cli.py` — 9 pass
- [x] `uv run pytest --doctest-modules src/ tests/ --ignore=tests/grammar/test_cli.py` — 63 pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,12 @@ dependencies = [
   "meds_testing_helpers~=0.3.0",
   "MEDS-transforms~=0.5.2",
   "meds-torch-data[lightning]~=0.7.0",
-  "transformers",
+  # Lower-bounded for the ``dtype=`` kwarg on ``AutoModelForCausalLM.from_config`` (4.56
+  # renamed ``torch_dtype`` → ``dtype`` and started deprecating the old name). ``head_dim`` as
+  # an explicit ``LlamaConfig`` attribute also landed in this era; the post-override snap
+  # logic in ``Model.__init__`` depends on it. Pinning a minimum avoids a class of obscure
+  # runtime failures on old installations.
+  "transformers>=4.56",
   "torch",
   "torchmetrics",
   "lightning~=2.5.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,11 +30,6 @@ dependencies = [
   "meds_testing_helpers~=0.3.0",
   "MEDS-transforms~=0.5.2",
   "meds-torch-data[lightning]~=0.7.0",
-  # Lower-bounded for the ``dtype=`` kwarg on ``AutoModelForCausalLM.from_config`` (4.56
-  # renamed ``torch_dtype`` → ``dtype`` and started deprecating the old name). ``head_dim`` as
-  # an explicit ``LlamaConfig`` attribute also landed in this era; the post-override snap
-  # logic in ``Model.__init__`` depends on it. Pinning a minimum avoids a class of obscure
-  # runtime failures on old installations.
   "transformers>=4.56",
   "torch",
   "torchmetrics",
--- a/src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
+++ b/src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
@@ -9,9 +9,6 @@ gpt_kwargs:
   num_attention_heads: ???
   attention_head_dim: ???
   hidden_size: ${int_prod:${.num_attention_heads},${.attention_head_dim}}
-  # Llama uses SwiGLU (3 projections vs NeoX GELU's 2), so matching NeoX parameter count would
-  # need intermediate_size ≈ (8/3)*hidden_size. We keep 4× here to preserve the YAML formula;
-  # callers targeting strict NeoX-param-parity can override intermediate_size explicitly.
   intermediate_size: ${int_prod:${.hidden_size},4}
 
   # Regularization
--- a/src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
+++ b/src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
@@ -9,12 +9,13 @@ gpt_kwargs:
   num_attention_heads: ???
   attention_head_dim: ???
   hidden_size: ${int_prod:${.num_attention_heads},${.attention_head_dim}}
+  # Llama uses SwiGLU (3 projections vs NeoX GELU's 2), so matching NeoX parameter count would
+  # need intermediate_size ≈ (8/3)*hidden_size. We keep 4× here to preserve the YAML formula;
+  # callers targeting strict NeoX-param-parity can override intermediate_size explicitly.
   intermediate_size: ${int_prod:${.hidden_size},4}
 
   # Regularization
   attention_dropout: 0.0
-  hidden_dropout: 0.0
-  classifier_dropout: 0.1
 
   # Dictated by the data
 
--- a/src/MEDS_EIC_AR/model/model.py
+++ b/src/MEDS_EIC_AR/model/model.py
@@ -41,14 +41,11 @@ def _val(tensor: torch.Tensor) -> int | bool | float:
 
 
 class Model(torch.nn.Module):
-    """A decoder-only transformer (Llama-style) for pre-training an autoregressive, "everything-is-code"
-    model.
+    """A Llama-style decoder-only transformer for pre-training an autoregressive, "everything-is-code" model.
 
-    This model is a wrapper around the Hugging Face ``LlamaForCausalLM`` model to run it over MEDS-TorchData
-    batches. The earlier revision wrapped ``GPTNeoXForCausalLM``; see issue #108 for the rationale behind the
-    swap (RMSNorm + SwiGLU + full-dim RoPE, and first-class support in modern inference engines like SGLang
-    / vLLM). The rolling-generation loop and everything downstream of :attr:`HF_model` are
-    architecture-agnostic, so swapping the base is contained to :meth:`__init__`.
+    Wraps :class:`~transformers.LlamaForCausalLM` so it runs over MEDS-TorchData batches. The rolling-
+    generation loop and everything downstream of :attr:`HF_model` are architecture-agnostic, so the base
+    is contained to :meth:`__init__` — swapping to a different HF architecture is a one-spot change.
 
     Args:
         gpt_kwargs: A dictionary of keyword arguments to pass to the underlying HF config constructor (a
@@ -103,8 +100,7 @@ class Model(torch.nn.Module):
                grad_fn=<UnsafeViewBackward0>)
 
     The models parameters can be accessed in the normal way. The first named parameter is the token
-    embedding matrix — Llama names it ``model.embed_tokens.weight`` (nested under our ``HF_model``
-    wrapper), versus NeoX's ``gpt_neox.embed_in.weight``:
+    embedding matrix:
 
         >>> sample_param_name, sample_param = next(iter(model.named_parameters()))
         >>> print(f"{sample_param_name} ({sample_param.shape}): {sample_param}")
@@ -141,10 +137,10 @@ class Model(torch.nn.Module):
             ...
         ValueError: Config for HF model llama does not have attribute foobar
 
-    Llama-specific post-override semantics: ``num_key_value_heads`` and ``head_dim`` are recomputed
-    from the caller's ``hidden_size`` / ``num_attention_heads`` overrides unless the caller set them
-    explicitly. Default is plain MHA (``num_key_value_heads == num_attention_heads``) to match the
-    NeoX behavior this swap replaces:
+    Post-override snap: ``num_key_value_heads`` and ``head_dim`` are recomputed from the caller's
+    ``hidden_size`` / ``num_attention_heads`` unless the caller set them explicitly. The default is
+    plain MHA (``num_key_value_heads == num_attention_heads``) — matching what vanilla
+    :class:`~transformers.LlamaConfig` does when either field is left unset:
 
         >>> tiny = Model({
         ...     "num_hidden_layers": 2,
@@ -213,17 +209,12 @@ class Model(torch.nn.Module):
         self.HF_model_config: LlamaConfig = LlamaConfig()
 
         for key, val in gpt_kwargs.items():
+            # ``attention_head_dim`` is consumed by the YAML ``int_prod`` resolver to derive
+            # ``hidden_size`` (``hidden_size = num_attention_heads * attention_head_dim``). The
+            # config expresses the product form so HPO search over ``num_attention_heads`` and
+            # ``attention_head_dim`` independently can never produce a non-divisible
+            # ``hidden_size``. It isn't a real LlamaConfig attribute — skip it on its way in.
             if key == "attention_head_dim":
-                # Derived into ``hidden_size`` via the config's ``int_prod`` expression; Llama
-                # infers ``head_dim`` from ``hidden_size // num_attention_heads`` itself. Logging at
-                # debug level so a caller who passes this via the Python API (bypassing the YAML
-                # ``int_prod`` resolver) can see why it had no effect.
-                logger.debug(
-                    "Ignoring gpt_kwargs key 'attention_head_dim' — it is consumed by the YAML "
-                    "``int_prod`` resolver to derive ``hidden_size`` and has no direct LlamaConfig "
-                    "equivalent. Set ``head_dim`` directly if you want to override Llama's "
-                    "``hidden_size // num_attention_heads`` default."
-                )
                 continue
             if not hasattr(self.HF_model_config, key):
                 raise ValueError(
@@ -231,14 +222,14 @@ class Model(torch.nn.Module):
                 )
             setattr(self.HF_model_config, key, val)
 
-        # Llama's ``num_key_value_heads`` (GQA) and ``head_dim`` are set from vanilla defaults at
-        # ``LlamaConfig()`` construction and are *not* recomputed when the caller overrides
-        # ``hidden_size`` / ``num_attention_heads`` via ``setattr`` above. Snap both back to the
-        # values implied by the caller's overrides unless the caller passed them explicitly:
-        # plain MHA (``num_key_value_heads == num_attention_heads``) matches the NeoX behavior
-        # this swap replaces (issue #108), and ``head_dim = hidden_size // num_attention_heads``
-        # is the same implicit relation NeoX used. A caller who wants GQA simply sets
-        # ``num_key_value_heads`` explicitly in ``gpt_kwargs``.
+        # ``LlamaConfig()``'s default values for ``num_key_value_heads`` and ``head_dim`` come from
+        # the vanilla 7B config — they aren't recomputed from the ``hidden_size`` /
+        # ``num_attention_heads`` the caller just overrode. Snap them back to the values implied
+        # by the overrides unless the caller also set them explicitly. Plain MHA
+        # (``num_key_value_heads == num_attention_heads``) and
+        # ``head_dim == hidden_size // num_attention_heads`` are what vanilla Llama uses when the
+        # fields are left unset at config-construction time; this just preserves that after our
+        # post-hoc attribute mutation.
         if "num_key_value_heads" not in gpt_kwargs:
             self.HF_model_config.num_key_value_heads = self.HF_model_config.num_attention_heads
         if "head_dim" not in gpt_kwargs:
@@ -248,7 +239,10 @@ class Model(torch.nn.Module):
                 raise ValueError(
                     f"hidden_size ({hidden}) must be divisible by num_attention_heads ({heads}) to "
                     "derive ``head_dim`` implicitly. Pass ``head_dim`` explicitly in gpt_kwargs, or "
-                    "adjust ``hidden_size`` / ``num_attention_heads`` so the division is exact."
+                    "adjust ``hidden_size`` / ``num_attention_heads`` so the division is exact. "
+                    "Configs built via the ``attention_head_dim`` resolver at "
+                    "``configs/lightning_module/model/default.yaml`` cannot hit this — the product "
+                    "form makes divisibility impossible-by-construction."
                 )
             self.HF_model_config.head_dim = hidden // heads
 
@@ -437,9 +431,8 @@ class Model(torch.nn.Module):
             - The parameters are not inf.
 
         Examples:
-            Llama has no attention biases by default (``attention_bias=False``), so the NeoX-era
-            ``query_key_value.bias`` target doesn't exist — we corrupt a ``RMSNorm`` weight instead.
-            The parameter choice doesn't matter; any tensor with known ``numel`` works.
+            We corrupt a ``RMSNorm`` weight to exercise the warning path. The parameter choice
+            doesn't matter; any tensor with known ``numel`` works.
 
             >>> model = Model({
             ...     "num_hidden_layers": 2,
--- a/src/MEDS_EIC_AR/model/model.py
+++ b/src/MEDS_EIC_AR/model/model.py
@@ -44,8 +44,11 @@ class Model(torch.nn.Module):
     """A Llama-style decoder-only transformer for pre-training an autoregressive, "everything-is-code" model.
 
     Wraps :class:`~transformers.LlamaForCausalLM` so it runs over MEDS-TorchData batches. The rolling-
-    generation loop and everything downstream of :attr:`HF_model` are architecture-agnostic, so the base
-    is contained to :meth:`__init__` — swapping to a different HF architecture is a one-spot change.
+    generation loop and much of the code downstream of :attr:`HF_model` are architecture-agnostic, so
+    :meth:`__init__` is the primary integration point with the underlying HF model. Swapping to a
+    different HF architecture may still require updates elsewhere — architecture-specific config keys,
+    or parameter-name-based logic like the no-weight-decay grouping in
+    :meth:`MEICARModule._is_norm_bias_param`.
 
     Args:
         gpt_kwargs: A dictionary of keyword arguments to pass to the underlying HF config constructor (a
@@ -99,7 +102,7 @@ class Model(torch.nn.Module):
                dtype=torch.float16,
                grad_fn=<UnsafeViewBackward0>)
 
-    The models parameters can be accessed in the normal way. The first named parameter is the token
+    The model's parameters can be accessed in the normal way. The first named parameter is the token
     embedding matrix:
 
         >>> sample_param_name, sample_param = next(iter(model.named_parameters()))
--- a/src/MEDS_EIC_AR/model/model.py
+++ b/src/MEDS_EIC_AR/model/model.py
@@ -140,6 +140,53 @@ class Model(torch.nn.Module):
         Traceback (most recent call last):
             ...
         ValueError: Config for HF model llama does not have attribute foobar
+
+    Llama-specific post-override semantics: ``num_key_value_heads`` and ``head_dim`` are recomputed
+    from the caller's ``hidden_size`` / ``num_attention_heads`` overrides unless the caller set them
+    explicitly. Default is plain MHA (``num_key_value_heads == num_attention_heads``) to match the
+    NeoX behavior this swap replaces:
+
+        >>> tiny = Model({
+        ...     "num_hidden_layers": 2,
+        ...     "num_attention_heads": 2,
+        ...     "hidden_size": 4,
+        ...     "max_position_embeddings": 10,
+        ...     "vocab_size": 40,
+        ... }, precision="32-true")
+        >>> tiny.HF_model_config.num_key_value_heads
+        2
+        >>> tiny.HF_model_config.head_dim
+        2
+
+    A caller who wants GQA sets ``num_key_value_heads`` explicitly:
+
+        >>> gqa = Model({
+        ...     "num_hidden_layers": 2,
+        ...     "num_attention_heads": 4,
+        ...     "num_key_value_heads": 2,
+        ...     "hidden_size": 8,
+        ...     "max_position_embeddings": 10,
+        ...     "vocab_size": 40,
+        ... }, precision="32-true")
+        >>> gqa.HF_model_config.num_key_value_heads
+        2
+        >>> gqa.HF_model_config.head_dim
+        2
+
+    And a non-divisible ``hidden_size`` / ``num_attention_heads`` combination is rejected at
+    construction time rather than producing a silently-wrong ``head_dim`` that only surfaces as a
+    cryptic shape error on the first forward pass:
+
+        >>> Model({
+        ...     "num_hidden_layers": 2,
+        ...     "num_attention_heads": 3,
+        ...     "hidden_size": 4,
+        ...     "max_position_embeddings": 10,
+        ...     "vocab_size": 40,
+        ... })
+        Traceback (most recent call last):
+            ...
+        ValueError: hidden_size (4) must be divisible by num_attention_heads (3) to derive ...
     """
 
     HF_model_config: LlamaConfig
--- a/src/MEDS_EIC_AR/model/model.py
+++ b/src/MEDS_EIC_AR/model/model.py
@@ -6,11 +6,10 @@ import torch.nn.functional as F
 from meds_torchdata import MEDSTorchBatch
 from omegaconf import DictConfig, OmegaConf
 from transformers import (
-    AutoConfig,
     AutoModelForCausalLM,
     GenerationConfig,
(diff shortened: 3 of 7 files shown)


mmcdermott:

mmcdermott:
Thanks for the careful review. Addressing each suggestion:

**1. ``attention_head_dim`` silent swallow → ✓ addressed in c941025**
Added a ``logger.debug(...)`` explaining why the key was ignored. Debug (not warn) because the YAML ``int_prod`` resolver feeds this key on every normal run — surfacing at warn would be noise; debug stays silent in typical usage but appears when someone turns on verbose logging to diagnose.

**2. ``head_dim`` divisibility assertion → ✓ addressed in c941025**
Now raises ``ValueError`` with a clear message before the integer division if ``hidden_size % num_attention_heads != 0``. You're right that without this, the failure mode was a confusing downstream shape error on the first forward pass.

**3. ``AutoModelForCausalLM.from_config`` → ``LlamaForCausalLM(...)`` → pushing back**
Verified the direct ``LlamaForCausalLM(config, **kwargs)`` constructor does **not** accept ``dtype`` or ``attn_implementation`` (tested in this repo's pinned transformers):

```
>>> LlamaForCausalLM(c, dtype=torch.float16)
TypeError: LlamaForCausalLM.__init__() got an unexpected keyword argument 'dtype'
```

Those kwargs are handled by ``PreTrainedModel._from_config`` (what ``AutoModelForCausalLM.from_config`` dispatches to), not the base ``__init__``. Switching to direct construction would require separately loading the model, calling ``.to(dtype)``, and setting ``attn_implementation`` via the private API — strictly worse for clarity than the current single-call dispatch. The Auto→Llama type relationship holds: ``AutoModelForCausalLM.from_config(LlamaConfig(...))`` returns a ``LlamaForCausalLM`` instance, so the ``HF_model: LlamaForCausalLM`` annotation matches the runtime type.

CI on c941025 coming up; doctests re-confirmed green locally.

mmcdermott:
chatgpt comments:
My take: this PR is basically right, and I would not block the architecture swap itself.

The core change seems coherent: swapping the wrapped HF base from NeoX to Llama, fixing the no-weight-decay grouping for the final Llama norm, and keeping the rolling-generation path architecture-agnostic all make sense. I do not see a deep design error in the diff.

I’d leave two asks:

1. Fix the stale docstring in `_norm_bias_param_names`. It still says “three norm weights per layer,” but the doctest output shows two per layer plus the final `model.norm.weight`. That is tiny, but it is an actual incorrect statement in the code.

2. Add one direct regression test or doctest for the new post-override snap semantics. That logic is behaviorally important now, and I’d want an explicit assertion for:

   * `num_key_value_heads == num_attention_heads` by default,
   * `head_dim == hidden_size // num_attention_heads`,
   * the non-divisible `hidden_size % num_attention_heads != 0` error path.

My only substantive concern is version compatibility around `dtype`. Current Transformers docs are clearly moving toward `dtype`, and the main docs now show `from_pretrained(..., dtype=...)` plus `AutoConfig` accepting `dtype` for models instantiated from scratch. But older docs centered on `torch_dtype`, and older releases documented narrower `from_config` signatures; there was also behavior churn across 4.46→4.49 around dtype-related loading semantics. So the PR is directionally correct, but if this repo intends to support a broad range of Transformers versions, I would either pin a minimum Transformers version or add a small compatibility shim instead of assuming `dtype` everywhere. ([[Hugging Face](https://huggingface.co/docs/transformers/main/en/models)][1])

Paste-ready comment:

> I think this is basically the right change. The architecture swap itself looks coherent, and the important behavioral fix here is excluding Llama’s final `model.norm.weight` from weight decay.
>
> Two small things I’d still want before merge:
>
> 1. `_norm_bias_param_names` has a stale docstring: it says “three norm weights per layer,” but the doctest output shows two per layer plus the final `model.norm.weight`.
> 2. Please add one direct regression test/doctest for the new post-override snap logic: default `num_key_value_heads == num_attention_heads`, default `head_dim == hidden_size // num_attention_heads`, and the non-divisible error path.
>
> My only substantive concern is compatibility around the `torch_dtype` -> `dtype` change. That is the current HF direction, but if this package is meant to work across older `transformers` versions, I’d either pin a minimum version or add a compatibility shim. If you’re happy requiring recent Transformers, then I think this is mergeable after the docstring/test cleanup.

[1]: https://huggingface.co/docs/transformers/main/en/models "Loading models · Hugging Face"

src/MEDS_EIC_AR/training/module.py
```diff
@@ -283,36 +288,29 @@ def _is_norm_bias_param(n: str) -> bool:
...
+        return bool(re.search(r"(bias|(layer_?)?norm(\d*)\.weight)", n, re.IGNORECASE))
 
     def _norm_bias_param_names(self) -> Iterator[str]:
         """Yields the names of parameters corresponding to the bias and normalization layers.
 
         These parameters should not be subject to weight decay by the optimizer.
 
         Examples:
+            Llama has no biases by default (``attention_bias=False``, MLP/LM-head biases absent) and
+            uses RMSNorm, so the only matches are the three norm weights per layer plus the final
```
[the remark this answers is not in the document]
mmcdermott: Fixed in c3fa275. Docstring now reads: "two norm weights per layer (``input_layernorm.weight`` + ``post_attention_layernorm.weight``) plus the final ``model.norm.weight``". Matches the doctest output literally. Same point was independently raised by ChatGPT in the top-level comment thread.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -154,16 +163,49 @@ class Model(torch.nn.Module):
...
+            hidden = self.HF_model_config.hidden_size
+            heads = self.HF_model_config.num_attention_heads
+            if hidden % heads != 0:
+                raise ValueError(
+                    f"hidden_size ({hidden}) must be divisible by num_attention_heads ({heads}) to "
+                    "derive ``head_dim`` implicitly. Pass ``head_dim`` explicitly in gpt_kwargs, or "
+                    "adjust ``hidden_size`` / ``num_attention_heads`` so the division is exact."
+                )
+            self.HF_model_config.head_dim = hidden // heads
+
```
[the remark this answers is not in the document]
mmcdermott: Added in c3fa275. Three new executable doctests inside the ``Model`` class docstring:

1. Default-MHA snap: a tiny `Model({hidden_size=4, num_attention_heads=2, ...})` now asserts `HF_model_config.num_key_value_heads == 2` and `head_dim == 2`.
2. Explicit GQA opt-in: passing `num_key_value_heads=2` alongside `num_attention_heads=4, hidden_size=8` asserts the explicit value survives the snap pass.
3. Non-divisible error path: `num_attention_heads=3, hidden_size=4` raises the `ValueError` added in c941025, proving we fail fast at construction rather than on the first forward pass.

All three run as part of the standard `pytest --doctest-modules src/MEDS_EIC_AR/model/model.py` suite. ChatGPT flagged the same gap in its top-level review; the tests land the coverage both reviewers asked for.

mmcdermott:
All three ChatGPT points addressed in c3fa275:

**1. Stale `_norm_bias_param_names` docstring** — fixed; now reads "two norm weights per layer (`input_layernorm.weight` + `post_attention_layernorm.weight`) plus the final `model.norm.weight`," matching the doctest output literally.

**2. Snap-semantics regression test** — added three executable doctests inside the `Model` class docstring:
- Default-MHA snap: `HF_model_config.num_key_value_heads == num_attention_heads`.
- Explicit GQA opt-in: `num_key_value_heads` override survives the snap pass.
- Non-divisible `hidden_size / num_attention_heads` raises a clear `ValueError` at construction rather than producing a cryptic shape error on the first forward pass.

**3. `dtype` vs `torch_dtype` compatibility** — pinned `transformers>=4.56` in `pyproject.toml`. `dtype` was renamed from `torch_dtype` in 4.56 and is what `AutoModelForCausalLM.from_config` accepts as canonical; older installs would either warn noisily or fail on our usage. The `head_dim` attribute on `LlamaConfig` also landed in this era, which the snap logic depends on.

Went with the version pin over a compat shim because:
- The shim would have to detect kwarg name dynamically (`inspect.signature` on `_from_config`) and switch between them, which is more surface area for a small window of old-version support.
- The repo already pins several upstream deps tightly (`polars~=1.30.0`, `lightning~=2.5.1`, `meds~=0.4.0`), so adding a lower bound on transformers is in-character for this project.
- Users running modern transformers (≥4.56) get the clean path today; users on older versions get a clear "install a newer transformers" install-time error instead of a confusing runtime failure.

If you'd rather keep the supported-transformers window wider, I can swap the pin for a `torch_dtype` compat shim and wire it behind `inspect.signature` detection — flag me if you'd prefer that path.

mmcdermott:
comments from chatgpt: 
Yes — at this point I think it is mergeable.

The three things I would have asked for are now in:

* the stale `_norm_bias_param_names` wording is fixed,
* the new Llama snap semantics now have explicit doctest coverage,
* and the `dtype` compatibility concern is handled by pinning `transformers>=4.56` in `pyproject.toml`. 

That leaves me with no real blocker. The architecture swap still looks coherent, the weight-decay fix for Llama’s final norm is still the key behavioral correction, and the rolling-generation code still appears appropriately architecture-agnostic. The latest PR state also shows the follow-up responses and edits landed on the branch, plus Codecov is now happy on modified lines. 

The only remaining “choice,” not objection, is policy: pinning `transformers>=4.56` is reasonable here, especially given the repo already pins several upstream dependencies fairly tightly. If you wanted a wider support window, you could do a compat shim instead, but I would not require that before merge. 

So my answer now is: **LGTM, merge once you’re satisfied with the version-floor decision.**

src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
```diff
@@ -9,12 +9,13 @@ gpt_kwargs:
   num_attention_heads: ???
   attention_head_dim: ???
   hidden_size: ${int_prod:${.num_attention_heads},${.attention_head_dim}}
+  # Llama uses SwiGLU (3 projections vs NeoX GELU's 2), so matching NeoX parameter count would
```
mmcdermott: This comment is irrelevant and can be dropped. And, to be clear -- we are not trying to match parameter count. We want to respect our best guess at the right transformer defaults possible. If that isn't 4 x the hidden size, we should adjust. Do we need to change here?

src/MEDS_EIC_AR/training/module.py
```diff
@@ -268,6 +268,11 @@ def validation_step(self, batch):
     def _is_norm_bias_param(n: str) -> bool:
```
mmcdermott, line 268: Just checking -- do we still need this functionality at all here? Summarize why this was important initially, then assess if it is still needed for this transformer base, and let me know.

pyproject.toml
```diff
@@ -30,7 +30,12 @@ dependencies = [
   "meds_testing_helpers~=0.3.0",
   "MEDS-transforms~=0.5.2",
   "meds-torch-data[lightning]~=0.7.0",
-  "transformers",
+  # Lower-bounded for the ``dtype=`` kwarg on ``AutoModelForCausalLM.from_config`` (4.56
```
mmcdermott: don't need this comment. Always good to pin a version floor.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -85,24 +90,26 @@ class Model(torch.nn.Module):
...
-        tensor([[[ 2.0309e-02, ...,  1.9135e-02], ..., [ 1.3962e-02, ...,  1.6510e-02]],
+        tensor([[[ 2.5539e-03, ..., -3.0045e-02], ..., [-8.3694e-03, ...,  3.0487e-02]],
         <BLANKLINE>
-                [[ 2.0309e-02, ...,  1.9135e-02], ..., [ 1.4549e-02, ...,  1.6312e-02]]],
+                [[ 2.5539e-03, ..., -3.0045e-02], ..., [-9.0561e-03, ...,  3.5919e-02]]],
                dtype=torch.float16,
                grad_fn=<UnsafeViewBackward0>)
 
-    The models parameters can be accessed in the normal way.
+    The models parameters can be accessed in the normal way. The first named parameter is the token
```
mmcdermott: you don't need the clarification vs. gpt_neox here.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -125,16 +132,65 @@ class Model(torch.nn.Module):
...
+    error message is derived from the active HF config's ``model_type`` so it stays accurate if we ever
+    swap the base architecture again:
 
         >>> Model({"foobar": 2})
         Traceback (most recent call last):
             ...
-        ValueError: Config for HF model gpt-neox does not have attribute foobar
+        ValueError: Config for HF model llama does not have attribute foobar
+
+    Llama-specific post-override semantics: ``num_key_value_heads`` and ``head_dim`` are recomputed
```
mmcdermott: Our defaults (and changes to the model size config options) should be our best guess at what would make a great architecture for this problem -- not what gpt neo x parity would imply. Does anything warrant adjustments based on that?

src/MEDS_EIC_AR/model/model.py
```diff
@@ -154,16 +210,49 @@ class Model(torch.nn.Module):
     def __init__(self, gpt_kwargs: dict | DictConfig, precision: str = "32-true", do_demo: bool = False):
         super().__init__()
 
-        self.HF_model_config: GPTNeoXConfig = AutoConfig.from_pretrained("EleutherAI/gpt-neox-20b")
+        self.HF_model_config: LlamaConfig = LlamaConfig()
 
         for key, val in gpt_kwargs.items():
             if key == "attention_head_dim":
+                # Derived into ``hidden_size`` via the config's ``int_prod`` expression; Llama
```
mmcdermott: I think this was always how this worked? Do you need a new log statement here? The point of that parameter was to avoid stupid errors due to individisibility during hyperparameter tuning. Honestly, we should do something similar for GQA heads too.

src/MEDS_EIC_AR/model/model.py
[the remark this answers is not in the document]
mmcdermott: This may be changed by some of my comments.

src/MEDS_EIC_AR/configs/lightning_module/model/default.yaml
mmcdermott: Dropped in 6794715 — the comment framed `4×` as "preserving NeoX parameter count," which is the wrong framing per your big-picture point. Reality: `4×` is what vanilla `LlamaConfig` uses at small-model scale (Llama-3.2 1B has `intermediate_size=8192, hidden_size=2048 → 4×`), so it already aligns with a modern Llama recipe. Larger Llama variants use `~8/3×` or `~3.5×`; different ratios are design choices, not correctness issues. Kept `4×` with no comment since it's defensible as a good default; happy to revisit if you have a specific target ratio in mind for our scale.

src/MEDS_EIC_AR/training/module.py
mmcdermott: Summary (also folded into the rewritten docstring in 6794715):

**Why it was originally added**: standard AdamW practice — bias parameters and the gain parameters of normalization layers (LayerNorm / RMSNorm) have different geometry from linear weights, and applying weight decay to them pulls normalizations toward collapse. The original AdamW paper, GPT-2, GPT-3, and GPT-NeoX all group these into a no-decay set. Lightning's `configure_optimizers` here builds two param groups and puts everything this predicate matches into the no-decay one.

**Is it still needed under Llama?** Yes, for the norm side. Llama uses RMSNorm (`input_layernorm.weight`, `post_attention_layernorm.weight`, and the final `model.norm.weight`) — all of which are gain parameters that still shouldn't receive weight decay for the same reason LayerNorm params shouldn't. The bias branch of the regex, however, is dormant under Llama's default (`attention_bias=False`, MLP bias off, LM-head bias off — Llama has no biases at all by default). I kept the bias branch in the regex as cheap insurance: anyone overriding `attention_bias=True` (valid `LlamaConfig` kwarg) or swapping to a bias-enabled architecture would otherwise silently weight-decay their biases. Zero performance cost when the bias branch never matches.

**Net**: keep the function, keep the regex as-is, rewritten docstring explains the rationale. The original `norm.weight` extension I made (dropping the `layer_` prefix requirement so Llama's final `model.norm.weight` gets caught) remains necessary — that's the one branch that genuinely needed updating for the Llama swap.

pyproject.toml
mmcdermott: Dropped the comment in 6794715; kept the `transformers>=4.56` pin.

src/MEDS_EIC_AR/model/model.py
mmcdermott: Removed in 6794715.

src/MEDS_EIC_AR/model/model.py
mmcdermott: Audited in 6794715. Current defaults and my read on each:

- **`num_key_value_heads == num_attention_heads`** (plain MHA): this is what vanilla `LlamaConfig()` does when `num_key_value_heads` is left unset. Modern Llama small-model recipes (Llama-3.2 1B) also use plain MHA. GQA is a deliberate opt-in — expose it by setting the key explicitly.
- **`head_dim == hidden_size // num_attention_heads`**: same — this is what `LlamaConfig()` does natively when `head_dim` is unset. Keeping it preserves the natural Llama derivation after our post-hoc `setattr` overrides.
- **`intermediate_size = 4 × hidden_size`** (`default.yaml`): matches what Llama-3.2 1B uses at our scale. Rewrote the comment to drop the NeoX framing.
- **`attention_dropout = 0.0`**: default.
- **`attn_implementation="flash_attention_2"` when `flash_attn` is installed**: Llama's HF implementation has a first-class FlashAttention 2 path.

Docstrings now frame these as "Llama defaults" — not "matching NeoX behavior."

src/MEDS_EIC_AR/model/model.py
mmcdermott: Right — the `hidden_size = num_attention_heads × attention_head_dim` product form in `default.yaml` was always the HPO-safety design: by construction, no HPO sweep can produce a non-divisible combination. Addressed in 6794715:

- Dropped the extraneous debug log on the `attention_head_dim` key swallow. Added a brief inline comment explaining why the YAML uses the product form, so a future reader understands it's an HPO-safety construct, not a compatibility hack.
- Kept the `hidden_size % num_attention_heads != 0` check in `Model.__init__` as insurance for Python-API callers who bypass the YAML resolver and pass a raw `hidden_size` — but the new error message now explicitly notes that configs built through the default YAML resolver can't trigger it. Failing fast at construction for direct callers is still a real win over a cryptic shape error on the first forward pass.

**Same pattern for GQA heads**: good idea, filed as a separate issue — #121. Three design options in there (product-form config like we do for head_dim, a `num_kv_groups` helper, or just add a fail-fast assertion parallel to the `head_dim` one). Left my recommendation there; not blocking this PR since nothing in the codebase actually exercises GQA today (plain MHA default).

src/MEDS_EIC_AR/model/model.py
```diff
@@ -42,14 +41,17 @@ def _val(tensor: torch.Tensor) -> int | bool | float:
...
 
 class Model(torch.nn.Module):
-    """A basic GPT-NeoX like model for pre-training an autoregressive, "everything-is-code" model.
+    """A Llama-style decoder-only transformer for pre-training an autoregressive, "everything-is-code" model.
 
-    This model is a wrapper around the Hugging Face GPTNeoXForCausalLM model to run it over MEDS-TorchData
-    batches.
+    Wraps :class:`~transformers.LlamaForCausalLM` so it runs over MEDS-TorchData batches. The rolling-
+    generation loop and everything downstream of :attr:`HF_model` are architecture-agnostic, so the base
+    is contained to :meth:`__init__` — swapping to a different HF architecture is a one-spot change.
```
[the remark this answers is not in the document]
mmcdermott: Good catch — you're right, this PR itself proves the claim was overstated (the ``MEICARModule._is_norm_bias_param`` regex widening + config key drops are outside ``__init__``). Softened in 1768586 to: "``__init__`` is the primary integration point ... swapping may still require updates elsewhere — architecture-specific config keys, or parameter-name-based logic like the no-weight-decay grouping in ``MEICARModule._is_norm_bias_param``." The cross-reference gives future maintainers a concrete example of the kind of extra-``__init__`` surface to audit.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -85,24 +87,25 @@ class Model(torch.nn.Module):
...
-        tensor([[[ 2.0309e-02, ...,  1.9135e-02], ..., [ 1.3962e-02, ...,  1.6510e-02]],
+        tensor([[[ 2.5539e-03, ..., -3.0045e-02], ..., [-8.3694e-03, ...,  3.0487e-02]],
         <BLANKLINE>
-                [[ 2.0309e-02, ...,  1.9135e-02], ..., [ 1.4549e-02, ...,  1.6312e-02]]],
+                [[ 2.5539e-03, ..., -3.0045e-02], ..., [-9.0561e-03, ...,  3.5919e-02]]],
                dtype=torch.float16,
                grad_fn=<UnsafeViewBackward0>)
 
-    The models parameters can be accessed in the normal way.
+    The models parameters can be accessed in the normal way. The first named parameter is the token
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 1768586.

PR #123, opened 2026-04-19
fix: preprocessing/__main__.py cleanup (closes #112, #113, #114, #115)

### LINKED ISSUE #112: preprocessing/__main__.py: buffers full subprocess stdout/stderr in memory for long-running stages
## What I found

`src/MEDS_EIC_AR/preprocessing/__main__.py` orchestrates `MEDS_transform-pipeline` and `MTD_preprocess` as subprocesses, with both calls using `subprocess.run(..., capture_output=True, check=False)`:

https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/preprocessing/__main__.py

`capture_output=True` buffers the full child stdout/stderr in-process until the child exits.  For small/demo cohorts this is fine, but for realistic runs it has two drawbacks:

1. **Memory**: a pipeline that emits per-shard progress over thousands of shards (normal for `MEDS_transform-pipeline` at MIMIC-IV scale) can produce hundreds of MB of stdout that sits in the parent Python process until the child exits.
2. **No liveness signal**: the user sees nothing from the orchestrator for however long the stage takes. If a stage hangs or stalls, it is indistinguishable from "working" until the overall timeout.

Noted while porting this file verbatim to [payalchandak/EveryQuery#74](https://github.com/payalchandak/EveryQuery/pull/74) — Copilot flagged the same thing there.

## Options

1. **Drop `capture_output=True`** (let the child inherit parent stdout/stderr). Live output, no buffering. Trade-off: `result.stdout` / `result.stderr` come back as `None`, so the error path can no longer dump them directly. Mitigated by the fact that both wrapped tools are Hydra-based and already write logs under `${hydra.run.dir}` — the error path can point the user there.
2. **Tee to a log file via `Popen`**: `Popen(cmd, stdout=open(log_fp, \"w\"), stderr=subprocess.STDOUT)`.  Live-to-disk, no RAM buffering, log file persists for debug.  Slightly more code.
3. **Status quo + size cap**: not directly supported; would need a manual `Popen` read loop.

I'd lean toward **option 2** — the log file is a concrete artifact the parent's error message can cite (\"MEDS_transform-pipeline failed, see `<log_fp>`\"), which is more useful than re-emitting captured bytes.

## Related

- Inconsistency in the error-logging path (MTD_preprocess block drops stdout on failure) — filing separately.
- Code-style cleanups (`copy.deepcopy(os.environ)` → `os.environ.copy()`) — filing separately.

Downstream: a fix here should propagate to [payalchandak/EveryQuery](https://github.com/payalchandak/EveryQuery) which now carries a byte-identical port (see [#74](https://github.com/payalchandak/EveryQuery/pull/74)).

### LINKED ISSUE #113: preprocessing/__main__.py: MTD_preprocess error path drops stdout; asymmetric with MEDS_transform block
## What I found

`src/MEDS_EIC_AR/preprocessing/__main__.py` has two subprocess calls with inconsistent error-logging:

```python
# MEDS_transform-pipeline block — logs BOTH stdout and stderr
if result.returncode != 0:
    logger.error("Error running MEDS_transform-pipeline")
    logger.error(result.stdout.decode())
    logger.error(result.stderr.decode())
    raise RuntimeError("Error running MEDS_transform-pipeline")

# MTD_preprocess block — logs ONLY stderr, drops stdout entirely
if result.returncode != 0:
    logger.error("Error running MTD_preprocess")
    logger.error(result.stderr.decode())
    raise RuntimeError("Error running MTD_preprocess")
```

For tools that emit diagnostic info on stdout (tracebacks interleaved with `print()` calls, progress lines before a hang, etc.), the MTD_preprocess block silently discards that. The inconsistency also makes the two paths feel subtly different to debug.

Additionally, the `RuntimeError` messages themselves don't carry the command or return code, so the exception text alone doesn't tell you which invocation failed or with what exit status — you have to scroll the logs.

Noted while porting this file verbatim to [payalchandak/EveryQuery#74](https://github.com/payalchandak/EveryQuery/pull/74) — Copilot flagged it there.

## Proposed fix

1. Symmetrize: log both `stdout` and `stderr` in the MTD_preprocess block (matching the MEDS_transform-pipeline block).
2. Include `cmd` + `returncode` in the `RuntimeError` message so a caller catching it has enough context without trawling the log stream.

Small patch, no behavior change besides better error visibility.

## Related

- #112 (capture_output buffering) — if that lands via streaming-to-log-file, this issue's fix would become "point the error at the log file path" rather than "log both streams". Probably worth resolving #112 first, then this one on top.

Downstream: fix should propagate to [payalchandak/EveryQuery](https://github.com/payalchandak/EveryQuery) which carries a byte-identical port (see [#74](https://github.com/payalchandak/EveryQuery/pull/74)).

### LINKED ISSUE #114: preprocessing/__main__.py: use os.environ.copy() instead of copy.deepcopy(os.environ); drop redundant copy in MTD_preprocess block
## What I found

Two small code-quality points in `src/MEDS_EIC_AR/preprocessing/__main__.py`:

### 1. `copy.deepcopy(os.environ)` is overkill

```python
env = copy.deepcopy(os.environ)
env["RAW_MEDS_DIR"] = str(input_dir)
```

`os.environ` is `str → str`; there are no nested mutable values that would benefit from a deep copy. `os.environ.copy()` (or `dict(os.environ)`) is the idiomatic pattern and drops the `import copy` dependency.

### 2. The MTD_preprocess block's env copy is dead code

```python
env = copy.deepcopy(os.environ)  # created but never mutated...

cmd = ["MTD_preprocess", ...]
result = subprocess.run(cmd, env=env, capture_output=True, check=False)
```

`env` is created, never mutated, then passed back to the child.  That's functionally equivalent to `env=os.environ` (or omitting `env=` entirely, which lets the child inherit the parent's env by default).

## Proposed fix

Patch:

```diff
-import copy
 import logging
 import os
 import subprocess
 ...
-        env = copy.deepcopy(os.environ)
+        env = os.environ.copy()
         env["RAW_MEDS_DIR"] = str(input_dir)
         ...
         result = subprocess.run(cmd, env=env, capture_output=True, check=False)
 ...
 # MTD_preprocess block — env unused, drop it
-        env = copy.deepcopy(os.environ)
-
         cmd = [...]
-        result = subprocess.run(cmd, env=env, capture_output=True, check=False)
+        result = subprocess.run(cmd, capture_output=True, check=False)
```

Flagged by Copilot on [payalchandak/EveryQuery#74](https://github.com/payalchandak/EveryQuery/pull/74#discussion_r3102078261) where this file was ported verbatim.

## Related

- #112 (capture_output)
- #113 or whichever number the error-logging-symmetry issue got

Downstream: fix should propagate to [payalchandak/EveryQuery](https://github.com/payalchandak/EveryQuery) which carries a byte-identical port.

### LINKED ISSUE #115: preprocessing/__main__.py: missing `if __name__ == "__main__"` guard makes `python -m ... preprocessing` silently no-op
## What I found

[`src/MEDS_EIC_AR/preprocessing/__main__.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/preprocessing/__main__.py) defines `process_data(cfg: DictConfig)` decorated with `@hydra.main(...)` but has no `if __name__ == "__main__": process_data()` guard at the end.

Result: `python -m MEDS_EIC_AR.preprocessing` silently does nothing — the module executes top-to-bottom (imports, decorator attaches, function defined) and then exits. The setuptools entry point `MEICAR_process_data = "MEDS_EIC_AR.preprocessing.__main__:process_data"` is unaffected (the wrapper calls the function explicitly), so `MEICAR_process_data --help` works fine. But `python -m MEDS_EIC_AR.preprocessing` — a reasonable thing to reach for while debugging or running without a wheel installed — is a silent no-op.

## Proposed fix

```python
if __name__ == "__main__":
    process_data()
```

at the end of the file. Same pattern the other Hydra-wrapped mains in MEICAR already use. Two-line change, zero risk.

## Context

Copilot flagged this on the [EveryQuery port](https://github.com/payalchandak/EveryQuery/pull/74) of this file. Fixed downstream in [payalchandak/EveryQuery 7324e78](https://github.com/payalchandak/EveryQuery/pull/74/commits/7324e78); filing here to keep MEICAR and EQ from diverging silently on this particular line.

Continues the small-cleanup trail from [#112](https://github.com/mmcdermott/MEDS_EIC_AR/issues/112), [#113](https://github.com/mmcdermott/MEDS_EIC_AR/issues/113), [#114](https://github.com/mmcdermott/MEDS_EIC_AR/issues/114).

mmcdermott:
## Summary

Four small preprocessing-entrypoint issues, addressed together because they all touch the same ~50 lines in `src/MEDS_EIC_AR/preprocessing/__main__.py`:

- **Closes #112** — long-running stages no longer buffer all stdout/stderr in memory. Switched `subprocess.run(..., capture_output=True)` to a `_run_streamed` helper that pipes the child's output straight through to the parent's streams. Live progress visibility on multi-minute runs; parent memory stays flat.
- **Closes #113** — MTD_preprocess error path no longer drops stdout. The streaming approach from #112 makes the stage-specific error-log blocks obsolete (both stages' output is already visible on failure). Replaced with a uniform "check the streamed output above" message from `_run_streamed`.
- **Closes #114** — `copy.deepcopy(os.environ)` → `os.environ.copy()` (environ is `str → str`, no nested mutables). Dropped the redundant env copy entirely in the MTD_preprocess block since it never mutated the env.
- **Closes #115** — added `if __name__ == "__main__"` guard so `python -m MEDS_EIC_AR.preprocessing` runs the entrypoint instead of silently no-op'ing. Console-script callers from `pyproject.toml` were unaffected; `python -m` invocation is a standard pattern users expect to work.

Dropped the now-unused `copy` import.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/grammar/test_cli.py` — 33 pass
- [x] Manual read of the refactored file; streaming / env / guard semantics are load-bearing for real-dataset runs but aren't covered by doctests. A real `MIMIC-IV` preprocessing run is the meaningful smoke test and happens outside CI.

### DIFF
--- a/src/MEDS_EIC_AR/preprocessing/__main__.py
+++ b/src/MEDS_EIC_AR/preprocessing/__main__.py
@@ -1,7 +1,7 @@
-import copy
 import logging
 import os
 import subprocess
+import sys
 from importlib.resources import files
 from pathlib import Path
 
@@ -11,6 +11,35 @@ from omegaconf import DictConfig
 logger = logging.getLogger(__name__)
 
 CONFIGS = files("MEDS_EIC_AR") / "preprocessing" / "configs"
+
+
+def _run_streamed(cmd: list[str], *, env: dict[str, str] | None = None, stage_name: str) -> None:
+    """Run a subprocess, streaming its output live to the parent's stdout/stderr.
+
+    The two MTD stages (``MEDS_transform-pipeline`` and ``MTD_preprocess``) can run for many
+    minutes on real datasets. Previously we used ``subprocess.run(..., capture_output=True)``
+    which buffered the entire stdout/stderr in memory until the child exited — on a several-
+    hour MIMIC-IV preprocessing job this meant zero progress visibility for the user and
+    unbounded parent-process memory growth. Piping straight through to the parent's streams
+    gives real-time progress and keeps memory flat at the cost of only surfacing output as
+    lines are produced (which is what you want for a user-facing CLI).
+
+    Also fixes the asymmetric error-logging pattern the pre-refactor code had: the old
+    MEDS_transform block logged both stdout and stderr on failure, but the MTD_preprocess
+    block only logged stderr, losing Python traceback context. Streaming both to the
+    parent's streams removes that asymmetry entirely — failure output is already visible.
+    """
+    logger.info(f"Running command: {' '.join(cmd)}")
+    # ``stdout`` / ``stderr`` defaulting to ``None`` in subprocess.run means "inherit the
+    # parent's file descriptors," which is the streaming behavior we want. Explicitly naming
+    # ``sys.stdout`` / ``sys.stderr`` so this stays correct under Hydra loggers that have
+    # redirected Python's streams (the underlying OS fd is still the right destination).
+    result = subprocess.run(cmd, env=env, stdout=sys.stdout, stderr=sys.stderr, check=False)
+    if result.returncode != 0:  # pragma: no cover
+        raise RuntimeError(
+            f"{stage_name} failed with exit code {result.returncode}. "
+            "Check the streamed output above for details."
+        )
 
 
 @hydra.main(version_base=None, config_path=str(CONFIGS), config_name="_process_data")
@@ -26,7 +55,10 @@ def process_data(cfg: DictConfig):
     if done_fp.exists():  # pragma: no cover
         logger.info("Pre-MTD pre-processing already done, skipping")
     else:
-        env = copy.deepcopy(os.environ)
+        # ``os.environ`` is ``str -> str``; no nested-mutable values to worry about, so a
+        # shallow copy is sufficient. ``os.environ.copy()`` returns a plain ``dict`` we're
+        # free to mutate without affecting the parent's environment.
+        env = os.environ.copy()
         env["RAW_MEDS_DIR"] = str(input_dir)
         env["MTD_INPUT_DIR"] = str(intermediate_dir)
 
@@ -35,18 +67,11 @@ def process_data(cfg: DictConfig):
             env["MIN_EVENTS_PER_SUBJECT"] = "1"
 
         pipeline_config_fp = (CONFIGS / "_reshard_data.yaml") if cfg.do_reshard else (CONFIGS / "_data.yaml")
-        cmd = [
-            "MEDS_transform-pipeline",
-            f"pipeline_config_fp={pipeline_config_fp!s}",
-        ]
-        logger.info(f"Running command: {' '.join(cmd)}")
-
-        result = subprocess.run(cmd, env=env, capture_output=True, check=False)
-        if result.returncode != 0:  # pragma: no cover
-            logger.error("Error running MEDS_transform-pipeline")
-            logger.error(result.stdout.decode())
-            logger.error(result.stderr.decode())
-            raise RuntimeError("Error running MEDS_transform-pipeline")
+        _run_streamed(
+            ["MEDS_transform-pipeline", f"pipeline_config_fp={pipeline_config_fp!s}"],
+            env=env,
+            stage_name="MEDS_transform-pipeline",
+        )
 
         logger.info("Pre-MTD pre-processing done")
         done_fp.touch()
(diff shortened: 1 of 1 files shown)


mmcdermott:

PR #124, opened 2026-04-19
chore: trim training-only callbacks from generation bundle (closes #120)

### LINKED ISSUE #120: Follow-up to #73: trim training-only callbacks from generation trainer bundle
## Motivation

`configs/trainer/callbacks/generation.yaml` (landed in PR #73) sets its defaults to `[default, generation_speed_logger, _self_]`. `default` in turn includes `model_checkpoint`, `early_stopping`, and `learning_rate_monitor` — all of which are training-lifecycle callbacks with `on_train_*` / `on_validation_*` hooks that never fire under `trainer.predict()`. They're inert but carried along.

ChatGPT flagged this as a non-blocking follow-up during PR #73 review: inert-but-muddy. Harmless in practice, but the abstraction boundary is wrong — a generation trainer pulling in a `ModelCheckpoint` implies something about the run that isn't true. Readers looking at the config have to know "these do nothing under predict" which is a footgun for anyone copy-pasting to a similar context (e.g., adding a `score_trajectories` entry point per #95).

### A. Replace `default` with an explicit minimum set

```yaml
# configs/trainer/callbacks/generation.yaml
defaults:
  - generation_speed_logger
  - _self_
```

Drops model_checkpoint / early_stopping / learning_rate_monitor from generation trainers entirely. The most honest configuration.

- Pro: reads as "generation runs need only the speed logger." No implicit deadweight.
- Con: if anyone was relying on one of the training callbacks appearing during generation (unlikely — they don't fire — but possible), this silently removes it.

### B. Factor out a `runtime_common` callback set

Create `configs/trainer/callbacks/runtime_common.yaml` with callbacks that are genuinely shared across train + predict (today there are none — this is speculative for future additions like system-metrics logging). Both `default` and `generation` depend on `runtime_common`. If/when we add shared callbacks, they land in the common file.

- Pro: future-proof.
- Con: empty-file today; over-engineered.

### C. Leave as-is, document the inertness

Add a comment to `generation.yaml` explaining why `default` is inherited and what the inheritance costs.

- Pro: zero risk.
- Con: documentation-only. Doesn't actually clean up the configuration surface.

## Recommendation

Option A. The training callbacks are inert under predict, the only callbacks the generation trainer actually wants are the generation-specific ones, and the fix is a one-line config change. Option B is premature; Option C leaves the footgun intact.

## Scope

In:
- `generation.yaml` defaults list trimmed to just the generation-specific callbacks.
- A terse `configs/README.md` note explaining the shape.
- Sanity-check grammar CLI test still passes (it doesn't depend on any training callback firing under predict).

Out:
- Any change to `default.yaml` itself (still used by pretrain).
- Restructuring the `demo_generate.yaml` / `generate.yaml` trainers.

cc #73, #53, #95

mmcdermott:
## Summary

**Closes #120.** The generation trainer's callback set was inheriting `model_checkpoint` / `early_stopping` / `learning_rate_monitor` from `default.yaml` — all training-lifecycle callbacks whose hooks never fire under `trainer.predict()`. They were inert but visibly present in the config. Dropped `default` from `configs/trainer/callbacks/generation.yaml`'s `defaults` list; kept only `generation_speed_logger`.

ChatGPT called this out as a non-blocking follow-up during PR #73 review. It's non-blocking because the callbacks are dead code under predict, not a correctness issue — but the muddy abstraction is a footgun for anyone copy-pasting from here (a future `score_trajectories` entrypoint would naturally start from this set and inherit a checkpoint callback it can't use).

## Test plan

- [x] \`uv run pytest tests/ --ignore=tests/grammar/test_cli.py\` — 33 pass. No test references the training callbacks from a predict context.

### DIFF
--- a/src/MEDS_EIC_AR/configs/trainer/callbacks/generation.yaml
+++ b/src/MEDS_EIC_AR/configs/trainer/callbacks/generation.yaml
@@ -1,5 +1,10 @@
+# Generation-time callback set. Deliberately does NOT inherit ``default`` because that brings
+# in ``model_checkpoint`` / ``early_stopping`` / ``learning_rate_monitor`` — all training-
+# lifecycle callbacks that don't fire under ``trainer.predict()``. They were inert but visibly
+# present in the config, which misled readers copy-pasting from here into related scenarios
+# (e.g. a future ``score_trajectories`` entrypoint). Keeping only callbacks that actually do
+# something during predict: the speed logger.
 defaults:
-  - default
   - generation_speed_logger
   - _self_
 


mmcdermott:

PR #130, opened 2026-04-19
fix: let subprocess inherit parent fds in preprocessing CLI (refs #127 review)

mmcdermott:
## Summary

Small follow-up fix to PR #123's streaming refactor, catching a bug Copilot flagged on the release-candidate PR #127.

**Previous** (from #123):
\`\`\`python
result = subprocess.run(cmd, env=env, stdout=sys.stdout, stderr=sys.stderr, check=False)
\`\`\`

**Bug**: `sys.stdout` and `sys.stderr` are Python-level wrappers that can be swapped by:
- pytest's capture mode (where `sys.stdout` has no real `.fileno()`),
- Jupyter notebooks (which replace both with `OutStream`),
- Any logging library that redirects stdout.

In those environments, `subprocess.run` either raises on `.fileno()` lookup or redirects to an unintended destination (a pytest capture buffer, a notebook cell, etc.).

**Fix**: let `stdout` / `stderr` default to `None` in `subprocess.run`. That inherits the *parent process's OS-level* file descriptors, which is exactly the streaming behavior we wanted for the long-running MTD pipelines and is robust to whatever Python-level wrappers are in play.

Also dropped the now-unused `sys` import.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/grammar/test_cli.py` — 33 pass.
- [x] Change is semantically equivalent for normal interactive / shell invocations (where `sys.stdout.fileno()` would have returned the real parent fd anyway), but now also works under pytest capture / Jupyter / logging-wrapper contexts.

### DIFF
--- a/src/MEDS_EIC_AR/preprocessing/__main__.py
+++ b/src/MEDS_EIC_AR/preprocessing/__main__.py
@@ -1,7 +1,6 @@
 import logging
 import os
 import subprocess
-import sys
 from importlib.resources import files
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


mmcdermott:

PR #135, opened 2026-04-20
fix: stabilize Model doctest against global torch print-options state

mmcdermott:
## Summary

- Make the `Model` class doctest deterministic by explicitly setting `torch.set_printoptions(edgeitems=1, threshold=50)` inside the `Examples:` block, and restoring defaults at the end.
- Regenerate the three expected tensor-repr outputs (`outputs.logits`, `embed_tokens.weight`, `sample_param.grad`) under those options — the previous expected strings had a middle-dim-on-one-line format that PyTorch 2.9.1's `_tensor_str_with_formatter` cannot actually produce, so the doctest only passed by coincidence under certain collection orders.

## Why this flaked

`MEDSTorchBatch.__repr__` in `meds_torchdata` temporarily lowers `threshold`/`edgeitems` while formatting itself and then restores defaults. If another doctest in the same pytest worker happened to call `sample_batch.__repr__` before the `Model` doctest, the Model tensor prints came out abbreviated; in cold collection orders they came out unabbreviated. The previous expected output didn't match either mode consistently, producing sporadic CI failures across PRs #124, #127, and #129 (release candidate + env-snapshot + cadence-allow-on-resume branches all hit it at various points).

## Fix approach

Belt-and-braces: set the print options in the doctest itself so the expected output is independent of upstream state. Restore with `torch.set_printoptions(profile="default")` after the tensor prints so later tests aren't affected.

## Test plan

- [x] `uv run pytest --doctest-modules src/MEDS_EIC_AR/model/model.py -q` — all 8 pass
- [x] `uv run pytest --doctest-modules src/ -q` — all 41 pass
- [ ] CI green on this PR
- [ ] After merge: PR #127 (release candidate) and PR #129 (env snapshot) pick up the fix via merge-from-dev

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,6 +10,7 @@ from typing import Any
 from unittest.mock import MagicMock, Mock, patch
 
 import pytest
+import torch
 from meds_testing_helpers.dataset import MEDSDataset
 from meds_torchdata import MEDSPytorchDataset, MEDSTorchBatch, MEDSTorchDataConfig
 from torch.utils.data import DataLoader
@@ -280,6 +281,11 @@ def _setup_doctest_namespace(
     pytorch_dataset: MEDSPytorchDataset,
     pytorch_dataset_with_task: MEDSPytorchDataset,
 ):
+    # Pin torch's tensor-repr to a fixed abbreviation pattern so doctests that print tensors
+    # are stable regardless of prior global state. ``MEDSTorchBatch.__repr__`` in
+    # ``meds_torchdata`` temporarily lowers threshold/edgeitems and restores defaults, so
+    # collection-order-dependent state would otherwise leak between tests.
+    torch.set_printoptions(edgeitems=1, threshold=50)
     doctest_namespace.update(
         {
             "print_warnings": partial(print_warnings, caplog),
--- a/conftest.py
+++ b/conftest.py
@@ -307,3 +307,7 @@ def _setup_doctest_namespace(
             "pytorch_dataset_with_task": pytorch_dataset_with_task,
         }
     )
+    yield
+    # Restore torch's print options to defaults after each test so the pin does not leak
+    # beyond the pytest session.
+    torch.set_printoptions(profile="default")
--- a/src/MEDS_EIC_AR/model/model.py
+++ b/src/MEDS_EIC_AR/model/model.py
@@ -60,7 +60,16 @@ class Model(torch.nn.Module):
         ValueError: If the gpt_kwargs contains a key that is not supported.
 
     Examples:
+        Force torch's tensor-repr to abbreviate consistently for the rest of the examples —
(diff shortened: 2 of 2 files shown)


src/MEDS_EIC_AR/model/model.py
```diff
@@ -60,7 +60,16 @@ class Model(torch.nn.Module):
...
+        Force torch's tensor-repr to abbreviate consistently for the rest of the examples —
+        otherwise the expected outputs below depend on whatever ``threshold`` and ``edgeitems``
+        the global state happens to have. ``MEDSTorchBatch.__repr__`` in ``meds_torchdata``
+        temporarily sets smaller values and restores to default right after, so ordering-
+        dependent collection in CI can leave us with torch's unabbreviated-by-default print
+        behavior for tensors smaller than 1000 elements. The parameters below were chosen to
+        reproduce the abbreviation pattern these doctests were originally generated against.
+
         >>> _ = torch.manual_seed(0)
+        >>> torch.set_printoptions(edgeitems=1, threshold=50)
```
mmcdermott: Can you just yield this in the doctest namespace fixture, maybe, so it applies to all torch tests and doesn't need to be set manually? Then you can delete the explanation above, too.

mmcdermott: Done in fcd3c75 — moved torch.set_printoptions(edgeitems=1, threshold=50) into the autouse _setup_doctest_namespace fixture in conftest.py, dropped the explanatory paragraph and the manual set_printoptions / restore lines from the Model docstring. The only tensor-repr doctests in src/ outside model.py are scalar tensors (tensor(1.) etc.), which are unaffected by edgeitems/threshold — safe to pin globally. Full --doctest-modules suite green locally.

conftest.py
```diff
@@ -280,6 +281,11 @@ def _setup_doctest_namespace(
     pytorch_dataset: MEDSPytorchDataset,
     pytorch_dataset_with_task: MEDSPytorchDataset,
 ):
+    # Pin torch's tensor-repr to a fixed abbreviation pattern so doctests that print tensors
+    # are stable regardless of prior global state. ``MEDSTorchBatch.__repr__`` in
+    # ``meds_torchdata`` temporarily lowers threshold/edgeitems and restores defaults, so
+    # collection-order-dependent state would otherwise leak between tests.
+    torch.set_printoptions(edgeitems=1, threshold=50)
```
mmcdermott, line 288: Is there a way to do this in a context manager, so it is automatically unset when the pytest test suite is done executing?

mmcdermott: Done in 81ec155 — the autouse fixture now yields and runs torch.set_printoptions(profile="default") on teardown, so the pinned state is guaranteed to be released back to defaults at the end of the pytest session (and between tests).

Issue #142, opened 2026-04-22
Streaming generation predict to avoid accumulating all outputs in memory


mmcdermott:
Follow-up to Copilot review comment on PR #127 ([id 3108121350](https://github.com/mmcdermott/MEDS_EIC_AR/pull/127#discussion_r3108121350)).

## Problem

`MEICAR_generate_trajectories` (src/MEDS_EIC_AR/__main__.py:282-313) materializes the full list of per-batch predictions before writing any parquet:

```python
predictions = trainer.predict(model=M, dataloaders=expanded_loader)
# ... then iterate and demux ...
per_trajectory_batches: dict[int, list[torch.Tensor]] = {t: [] for t in range(n_trajectories)}
for pred in predictions:
    ...
```

With the N-sample interleaving from #103, the expanded dataset has `len(base_dataset) * n_trajectories` rows, so peak host memory during generation scales multiplicatively. On large real cohorts this is a real ceiling.

The companion fix in #<this-PR> moves the per-batch return tensors to CPU (fixing the GPU-memory half of the problem), but peak host RAM is still ~`len(base_dataset) * n_trajectories * max_new_tokens * sizeof(token)` plus bookkeeping.

## Proposed direction

Switch to a streaming model:

1. Call `trainer.predict(..., return_predictions=False)`.
2. Register a predict-time callback / writer that receives each batch's `(tokens, subject_idxs, trajectory_idxs)`, demuxes into the per-trajectory buckets in flight, and writes parquets as trajectories' subject ranges complete.

Because the expanded dataset yields rows in subject-index order within each trajectory, a trajectory's rows arrive in `format_trajectories`-compatible order. A streaming writer can either:
- buffer per-trajectory and flush at batch granularity (simple, still O(n_trajectories) open parquet writers), or
- buffer per-trajectory up to a row cap and write sharded parquets per trajectory+shard (bounded memory, adds a concat step on read).

The first is probably sufficient.

## Non-goals

- Not changing public CLI or the `generate_trajectories` output layout.
- Not changing `format_trajectories`'s contract — it still takes a list of per-batch token tensors.

Refs #127, #103, #88.

mmcdermott:
## Design analysis — scoping a minimal DDP-parallel fix

Reviewing the current plumbing before picking an approach:

[...]

PR #143, opened 2026-04-22
fix: pin generate trainer to devices=1 + assert demux ordering (refs #127)

mmcdermott:
## Summary

Address Copilot review comment on PR #127 ([id 3124878918](https://github.com/mmcdermott/MEDS_EIC_AR/pull/127#discussion_r3124878918)): the generation demux path assumes `trainer.predict(...)` returns per-batch outputs in the same global subject-index order as `base_dataset.schema_df`, because `format_trajectories` walks the schema df via a running `slice(st_i, batch_size)`. That assumption silently breaks under DDP / multi-device predict (Lightning stripes rows across ranks and collects per-rank outputs), yielding misaligned `(subject_id, prediction_time)` metadata in the written parquets.

## Changes

Two complementary guards:

1. **`configs/trainer/generate.yaml`** now pins `devices: 1`. The file previously inherited `devices: "auto"` from `default.yaml`, so a multi-GPU host would hit the bug by default at the top-level `MEICAR_generate_trajectories` entry point. Single-process predict is the only tested-correct configuration today.
2. **Defensive runtime check** in `__main__.py` verifies that the concatenated `subject_idxs` per trajectory arrive as `[0, 1, ..., len(base_dataset)-1]` before calling `format_trajectories`. Fails loudly if anyone removes the `devices: 1` pin, shuffles the expanded dataloader, or changes the `RepeatedPredictionDataset` ordering, instead of silently writing corrupt trajectory parquets.

DDP-parallel generation — which would use `subject_idxs` to explicitly reassemble outputs rather than rely on order-preservation — is the right long-term answer and is tracked in #142.

## Scope

Targeted at `dev` so it flows into release candidate #127 before the merge to `main`. The fix is localized to the two files above; no change to `format_trajectories`, `RepeatedPredictionDataset`, or the public CLI.

## Test plan

- [x] `uv run pytest tests/ -q -k generate` → 7/7 passed
- [ ] CI green on this PR

Refs #127.

### DIFF
--- a/src/MEDS_EIC_AR/__main__.py
+++ b/src/MEDS_EIC_AR/__main__.py
@@ -292,9 +292,11 @@ def generate_trajectories(cfg: DictConfig):
         # trajectories-per-subject that row corresponds to (0..N-1). It's distinct from
         # ``subject_idxs`` which records the base-dataset index the row came from.
         per_trajectory_batches: dict[int, list[torch.Tensor]] = {t: [] for t in range(n_trajectories)}
+        per_trajectory_subject_idxs: dict[int, list[torch.Tensor]] = {t: [] for t in range(n_trajectories)}
         for pred in predictions:
             tokens = pred["tokens"]
             trajectory_idxs = pred["trajectory_idxs"]
+            subject_idxs = pred["subject_idxs"]
             # Iterate over the trajectories actually present in this batch rather than always
             # doing N boolean compares. For batches that cover every trajectory (the common case
             # when batch_size >= N), this is the same work; for tail batches or small batch sizes,
@@ -302,6 +304,30 @@ def generate_trajectories(cfg: DictConfig):
             for t in trajectory_idxs.unique().tolist():
                 mask = trajectory_idxs == t
                 per_trajectory_batches[t].append(tokens[mask])
+                per_trajectory_subject_idxs[t].append(subject_idxs[mask])
+
+        # Defense against silent misalignment. ``format_trajectories`` walks
+        # ``base_dataset.schema_df`` via a running ``slice(st_i, batch_size)``, assuming the
+        # concatenated predictions for trajectory ``t`` arrive in subject-index order
+        # ``0..len(base_dataset)-1``. That assumption holds for the configured
+        # single-device predict path (``configs/trainer/generate.yaml`` pins ``devices: 1``),
+        # but any future move to DDP-parallel predict, any accidental shuffle, or any change
+        # to the expanded-dataset ordering would silently misalign the written (subject_id,
+        # prediction_time) metadata. Fail loud here instead. DDP-parallel generation is
+        # tracked at #142 and would replace this check with an explicit reassembly step.
+        n_subjects = len(base_dataset)
+        expected_order = torch.arange(n_subjects)
+        for t, subj_idxs_batches in per_trajectory_subject_idxs.items():
+            got = torch.cat(subj_idxs_batches) if subj_idxs_batches else torch.empty(0, dtype=torch.long)
+            if got.shape != expected_order.shape or not torch.equal(got, expected_order):
+                raise RuntimeError(
+                    f"Trajectory {t} predictions arrived out of expected subject-index order: "
+                    f"got {got.tolist()[:10]}... (len {got.shape[0]}); "
+                    f"expected [0, 1, ..., {n_subjects - 1}]. This can happen under multi-device "
+                    "predict (DDP stripes rows across ranks) or if the expanded dataloader is "
+                    "shuffled. Pin `trainer.devices=1` for generation; see issue #142 for the "
+                    "streaming-reassembly follow-up."
+                )
 
         for trajectory_idx, out_fp in trajectory_paths.items():
             if out_fp.is_file() and not cfg.do_overwrite:
--- a/src/MEDS_EIC_AR/__main__.py
+++ b/src/MEDS_EIC_AR/__main__.py
@@ -18,13 +18,11 @@ from importlib.resources import files
 from pathlib import Path
 
 import hydra
-import pyarrow.parquet as pq
 import torch
 from hydra.utils import instantiate
 from lightning.pytorch import seed_everything
 from meds import held_out_split, train_split, tuning_split
 from meds_torchdata import MEDSTorchBatch, MEDSTorchDataConfig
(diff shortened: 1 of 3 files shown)


src/MEDS_EIC_AR/__main__.py
```diff
@@ -292,16 +292,42 @@ def generate_trajectories(cfg: DictConfig):
...
             # Iterate over the trajectories actually present in this batch rather than always
             # doing N boolean compares. For batches that cover every trajectory (the common case
             # when batch_size >= N), this is the same work; for tail batches or small batch sizes,
             # it scales with the number of distinct trajectory_idxs in the batch instead.
             for t in trajectory_idxs.unique().tolist():
                 mask = trajectory_idxs == t
                 per_trajectory_batches[t].append(tokens[mask])
+                per_trajectory_subject_idxs[t].append(subject_idxs[mask])
+
+        # Defense against silent misalignment. ``format_trajectories`` walks
```
mmcdermott: Should we potentially split this function into a few internal functions -- one that reads through all the trajectories and combines them, one that validates the returned trajectories with this new error function and anything else relevant, and one that re-formats and writes them out as desired? That would allow us to properly test this, right?

mmcdermott (commented):
Will this impact only multi-node prediction, or any multi-GPU prediction? And how hard would it be to fix this properly, so that the generation process tracks sufficient information to be robust regardless of other settings?

src/MEDS_EIC_AR/__main__.py
mmcdermott: Yes, agreed — done in 9151b91.

Split into three helpers in a new ``MEDS_EIC_AR.generation.runner`` module (sibling of ``format_trajectories.py`` / ``repeated_dataset.py``):

1. ``demux_predictions_per_trajectory(predictions, n_trajectories)`` — fans ``Trainer.predict`` batches out into per-trajectory ``{tokens, subject_idxs}`` streams. Preserves per-batch grouping so variable-``L`` generator outputs do not need padding.
2. ``assert_expected_subject_index_order(per_trajectory_subject_idxs, n_subjects)`` — the defensive check.
3. ``write_per_trajectory_parquets(per_trajectory_batches, trajectory_paths, base_dataset, do_overwrite)`` — the format + per-file-idempotent write.

Each helper has a doctest block. The assertion helper in particular now has explicit coverage for every silent-failure mode: shuffled / DDP-gathered, short (missing rows), long (duplicate / oversample), empty per-trajectory, and a "second trajectory fails while the first passes" case that confirms the loop reports the right trajectory.

Ended up in `generation/runner.py` (not `__main__.py`) because pytest skips `__main__.py` from default doctest collection — moving the helpers to a sibling module lets `pytest --doctest-modules src/` pick them up without special-casing.

Local run: `uv run pytest --doctest-modules src/ -q` → 43 passed (was 41; +2 for the new doctest blocks). `uv run pytest tests/ -q -k "generate or grammar"` → 25 passed, confirms the full-CLI path still works.

mmcdermott:
Good questions — answers in order.

**Scope of the bug.** Any multi-GPU predict, not only multi-node. Lightning auto-selects a strategy based on `accelerator` and `devices`: one GPU → single-process (no DDP, bug not triggered); two or more GPUs on a single host → DDP auto-activates; multi-node → DDP across nodes. Under DDP, `DistributedSampler` stripes rows across ranks and `Trainer.predict` gathers per-rank outputs; the gathered list is ordered rank-major, not subject-index-order, which is exactly what silently misaligns the `format_trajectories.slice(st_i, B)` walk. So today, running `MEICAR_generate_trajectories` on a stock multi-GPU host with the default `devices: "auto"` would silently corrupt the output parquets before this PR.

**How hard to fix properly.** Medium, not small. The right fix is option (b) from Copilot (https://github.com/mmcdermott/MEDS_EIC_AR/pull/127#discussion_r3124878918): after gathering, reassemble outputs using the returned `subject_idxs` so ordering is explicit and the formatter no longer depends on arrival order.

The one complication is that each row has its own `L` (`model.generate` produces a variable-length sequence per row within a batch, padded to the batch max). Concat-across-batches to a single global tensor needs either:

1. Pad to global max-`L` — simple, wasteful on long tails.
2. Emit as list of 1-row "batches" in sorted subject order — correct with `format_trajectories` as-is, but one polars slice call per subject (O(N) polars overhead; probably fine in practice — polars slices are O(1) on arrow-backed frames).
3. Refactor `format_trajectory_batch` to take an explicit `(subject_id, prediction_time, last_time, tokens)` iterable — cleanest but biggest diff; contract change.

I estimate option (2) is ~30 lines plus doctests, and has no downstream contract change. It is the natural follow-up to the split in 9151b91 — the order-assertion helper would become an order-sort helper, taking the same input shape.

**Recommendation.** Ship the pin + assert as-is in this PR (release-blocking, keeps #127 correct), and track the proper DDP-parallel path in #142. Happy to pick up #142 next if you want me to do it before the #127 merge to `main` — the tradeoff is release-window time vs. generation-speed on large cohorts. The assert in this PR guarantees #127 cannot silently corrupt output; DDP-parallel generation is an optimization, not a correctness gap.

mmcdermott:
Agreed — done in 5c03e90.

New module shape in `generation/runner.py`:

- **`write_predictions(predictions, *, n_trajectories, trajectory_paths, base_dataset, do_overwrite)`** — single public entry point. Takes `Trainer.predict` output, writes one parquet per trajectory. This is what `__main__.py` now calls.
- **Private helpers** (`_demux_predictions_per_trajectory`, `_assert_expected_subject_index_order`, `_write_per_trajectory_parquets`) — implementation detail, underscore-prefixed to signal "do not import from outside this module". Their doctests still get collected by `pytest --doctest-modules src/` (the leading underscore only hides from `from module import *`, not from pytest discovery).

`__main__.generate_trajectories` now calls the composed endpoint in one line:

```python
write_predictions(
    predictions,
    n_trajectories=n_trajectories,
    trajectory_paths=trajectory_paths,
    base_dataset=base_dataset,
    do_overwrite=cfg.do_overwrite,
)
```

Local run: `--doctest-modules src/ -q` → 43 passed (unchanged — the three helpers still have their doctests, just under private names). `tests/ -k "generate or grammar"` → 25 passed in ~7.5 min (full CLI path still green through the collapsed single-endpoint composition).

PR #153, opened 2026-04-29
fix: persist + restore logger run ids correctly across interrupted training and generation resumes

### LINKED ISSUE #131: Generation logger restore/save path is asymmetric: restores from training, saves to generation (orphan write)
## The bug

`MEICAR_generate_trajectories` in `__main__.py::generate_trajectories` currently:

1. **Restores** logger run-ids from `cfg.model_initialization_dir` (the training run's output dir).
2. **Saves** its own logger run-ids into `cfg.output_dir` (the generation run's output dir).

The two halves don't match. Nothing ever reads the saved generation run-ids back — future generation resumes just re-read from `model_initialization_dir` again, which still has the training run's ids. The save is an orphan write.

Spotted by ChatGPT on PR #127 (release-candidate) review.

## Why it exists (history)

The inconsistency came from two rounds of PR #73 review:

- Initial state: both restore + save were against `cfg.model_initialization_dir`. Symmetric.
- Copilot flagged that saving generation ids into the training dir would overwrite the training run's save-point in the "fresh run_id for generation" escape-hatch case.
- I moved only the save to `cfg.output_dir` to fix that. Left the restore where it was because that's where the training run_id genuinely lives.
- Never closed the loop that now a future generation resume reads stale ids from the training dir and ignores the fresh ids written to the generation dir.

### A. Layered lookup: restore from output_dir first, fall back to model_init_dir

`apply_saved_logger_run_ids` checks `cfg.output_dir/loggers/mlflow_run_id.txt` first; if absent, falls back to `cfg.model_initialization_dir/loggers/mlflow_run_id.txt`.

- Pro: internally coherent. Generation saves → generation reads on resume. Training run is inherited on the first-ever generation run and then stays inherited unless the user overrides.
- Con: adds a second lookup path; need to decide the precedence carefully. Worth it since this IS how the system is supposed to work.

### B. Drop the save in generate_trajectories

If the intent is always "generation continues in the training run's mlflow state," never save generation run-ids — just inherit from training. This is simpler but loses the "fresh run_id for generation" escape hatch.

- Pro: minimal code change. One deletion.
- Con: the `run_id: fresh` explicit override no longer persists across resumes. Users who set `run_id=foo` once would have to re-pass it every time.

### C. Save to both

After generation, write the new run-id to both `output_dir` AND `model_initialization_dir`. Next generation resume reads from `model_init_dir`, finds the fresh id.

- Pro: simplest to implement.
- Con: clobbers the training run's save-point (which Copilot specifically asked us not to do). Regresses the earlier #73 review.

## Recommendation

**A** (layered lookup). It's the only option that handles both the "continue training's mlflow run by default" and "generation can have its own mlflow run that persists across resumes" cases without clobbering the training save-point.

## Tests needed

- `test_generate_trajectories_resumes_from_own_saved_run_id`: run `generate_trajectories` twice with the same `output_dir` and a fresh-run-id override on the first call; the second call should pick up the first call's saved run-id from `output_dir`, not the training run-id from `model_initialization_dir`.
- `test_generate_trajectories_falls_back_to_training_run_id_when_output_dir_empty`: running generation on a brand-new `output_dir` with no prior ids should restore from `model_initialization_dir`.

## Priority

Medium. Not a hard correctness bug (MLflow still receives logs, just from the wrong run depending on how you read "wrong"), but it's the kind of "looks right but does something subtly off" that will bite someone debugging why their two generation runs landed in different MLflow runs. Worth fixing before the next real training+generation sweep.

cc #53, #73, #127

### LINKED ISSUE #152: save_logger_run_ids fires only after trainer.fit() returns — wandb resume never works on the case it's designed for
### Bug
`save_logger_run_ids(trainer.loggers, output_dir)` is invoked **after** `trainer.fit(...)` returns at `src/MEDS_EIC_AR/__main__.py:152-153`:

```python
trainer.fit(**trainer_kwargs)
save_logger_run_ids(trainer.loggers, output_dir)
```

This means the run-id file (`<output_dir>/loggers/wandb_run_id.txt`) is only persisted on **clean completion** of `max_steps` — the exact case where wandb resume is *not* needed. For the interrupted-and-resumed case (OOM, SIGINT, machine crash, OS reboot — what the helper exists to support) `trainer.fit()` never returns cleanly, line 153 doesn't execute, and on the next `MEICAR_pretrain` invocation `apply_saved_logger_run_ids` finds no file and wandb spawns a fresh run.

The same shape exists for generation at `__main__.py:354`.

### Empirical repro
A current `MEICAR_pretrain` run (`v0.3.0`, ~13 h elapsed, ~14k steps) has been logging to a wandb run for hours but `models/v0.3.0/pretrain/loggers/wandb_run_id.txt` does not exist:

```
$ ls models/v0.3.0/pretrain/loggers/
wandb/
$ ls models/v0.3.0/pretrain/loggers/wandb/
run-20260427_174050-akdbm2fl
```

If this run dies before hitting `max_steps`, the user-visible behavior on resume is "wandb continuity lost" even though both `apply_saved_logger_run_ids` and `save_logger_run_ids` are wired up.

### Proposed fix
Persist the run id at **run-start**, not run-end. The cleanest path is a small Lightning Callback that calls `save_logger_run_ids` from `on_train_start` (or `setup`) — that's the first moment `WandbLogger.experiment` is materialized and `.id` is stable. The current end-of-fit call can stay (idempotent) for the clean-completion case, but it's the start-of-fit call that actually matters.

Sketch:

```python
class _PersistRunIDsOnStart(L.Callback):
    def __init__(self, output_dir: Path):
        self.output_dir = output_dir
    def on_train_start(self, trainer, pl_module):
        save_logger_run_ids(trainer.loggers, self.output_dir)

# in pretrain():
trainer = instantiate(cfg.trainer)
trainer.callbacks.append(_PersistRunIDsOnStart(output_dir))
trainer.fit(**trainer_kwargs)
```

Mlflow gets the same fix for free since `save_logger_run_ids` already handles both backends.

### Versions
- `MEDS-EIC-AR==0.3.0`
- Python 3.12, lightning 2.5.x, wandb 0.26.1, single GPU

mmcdermott:
## Summary

Two distinct bugs in the logger run-id save / restore flow, with **failing
regression tests landed first** so the fix can be reviewed against the
contract they pin down.

This PR is **draft** while the failing tests are evaluated. Once the test
shape is agreed, a follow-up commit will land the fix and flip the tests
green; only then will it be marked ready for review.

### Bug A — #152: ``save_logger_run_ids`` fires only after ``trainer.fit()`` returns

Today ``__main__.pretrain`` calls ``save_logger_run_ids`` *after* ``trainer.fit(...)``
returns. For a clean run that's fine; for the case the helper exists to
support (interrupted-and-resumed runs from OOM, SIGINT, OS reboot)
``trainer.fit`` never returns, the save line never executes, no
``wandb_run_id.txt`` is written, and the next ``MEICAR_pretrain do_resume=True``
invocation finds nothing to restore — wandb spawns a fresh run and continuity
is silently lost.

**Fix design** (intended for the follow-up commit on this PR): persist run
ids on ``on_train_start`` via a small Lightning Callback. The trailing
``save_logger_run_ids`` after ``trainer.fit`` can stay (idempotent for the
clean-completion case) but is no longer load-bearing.

**Failing test**: ``test_pretrain_persists_wandb_run_id_when_fit_interrupted``
patches ``Trainer.fit`` to fire ``on_train_start`` callbacks then raise
(simulating a crash), drives ``pretrain.__wrapped__(cfg)`` in-process with
an offline wandb logger, and asserts ``loggers/wandb_run_id.txt`` is on disk
when the exception propagates.

### Bug B — #131: ``generate_trajectories`` save / restore paths don't connect

``generate_trajectories`` *restores* run ids from
``cfg.model_initialization_dir`` but *saves* them to ``cfg.output_dir``.
Nothing reads ``output_dir`` back on the next generation resume — it always
re-restores from the training dir. The save is an orphan write.

**Fix design** (intended for the follow-up commit): layered lookup. Add a
``fallback_dir`` parameter to ``apply_saved_logger_run_ids``. Generation
calls it as ``apply_saved_logger_run_ids(cfg.trainer, output_dir,
fallback_dir=model_initialization_dir)``. A saved id under ``run_dir``
wins; if ``run_dir`` has no saved id, fall back to ``fallback_dir``. This
preserves the training save-point (no clobber) while making the generation
save-point actually load on the next resume.

**Failing tests**:
- ``test_apply_saved_logger_run_ids_prefers_run_dir_over_fallback`` — both
  dirs have files; ``run_dir`` wins.
- ``test_apply_saved_logger_run_ids_falls_back_when_run_dir_empty`` —
  ``run_dir`` empty, ``fallback_dir`` has files; fallback wins.

Both fail today because the helper has no ``fallback_dir`` parameter.

## Test plan

- [x] New tests fail locally for the right reasons (3 failures: 1 from a
      missing on-disk file, 2 from ``TypeError: unexpected keyword argument
      'fallback_dir'``).
- [ ] Confirm with the maintainer that the test shape pins down the
      intended contract before landing the fix commit.
- [ ] Follow-up commit lands the fix; rerun this test file expecting all
      green.
- [ ] Full ``uv run pytest`` green before marking ready for review.

Closes #152
Closes #131


src/MEDS_EIC_AR/training/callbacks.py
```diff
@@ -1,11 +1,37 @@
...
+from pathlib import Path
 
 from lightning.pytorch.callbacks import Callback
 from lightning.pytorch.loggers import Logger
 
 
+class SaveLoggerRunIDsOnTrainStart(Callback):
+    """Persist MLFlow / WandB run ids on ``on_train_start`` so an interrupted fit is resumable.
+
+    ``MEICAR_pretrain`` previously called :func:`MEDS_EIC_AR.utils.save_logger_run_ids`
```
mmcdermott: We don't need historical bug context here

src/MEDS_EIC_AR/training/callbacks.py
```diff
@@ -1,11 +1,37 @@
...
+    WandbLogger ``.experiment`` is initialized lazily during ``setup``), so reading
+    ``trainer.loggers[i].experiment.id`` is safe at this hook. Issue #152 has the full
+    motivation.
+    """
+
+    def __init__(self, output_dir: Path | str):
+        self.output_dir = Path(output_dir)
+
+    def on_train_start(self, trainer, pl_module):
+        from MEDS_EIC_AR.utils import save_logger_run_ids
```
mmcdermott: just do a top-level import. This import is not expensive. Unless there is another reason to do this?

src/MEDS_EIC_AR/training/callbacks.py
```diff
@@ -1,11 +1,37 @@
...
+    any compute that might crash, so resume sees the saved id even after a hard failure.
+
+    Lightning fires ``on_train_start`` after the loggers have been materialized (the
+    WandbLogger ``.experiment`` is initialized lazily during ``setup``), so reading
+    ``trainer.loggers[i].experiment.id`` is safe at this hook. Issue #152 has the full
+    motivation.
+    """
+
+    def __init__(self, output_dir: Path | str):
+        self.output_dir = Path(output_dir)
```
mmcdermott: Can this instead get a directory from the trainer? Doesn't the trainer know where things should be stored? Feel free to push back if I'm wrong.

src/MEDS_EIC_AR/__main__.py
```diff
@@ -135,6 +140,9 @@ def pretrain(cfg: DictConfig):
 
     apply_saved_logger_run_ids(cfg.trainer, output_dir)
     trainer = instantiate(cfg.trainer)
+    # Persist run ids on training start (not just on fit return) so an interrupted run is
+    # resumable. See issue #152.
```
mmcdermott: I don't think we need this comment.

src/MEDS_EIC_AR/__main__.py
```diff
@@ -135,6 +140,9 @@ def pretrain(cfg: DictConfig):
...
+    trainer.callbacks.append(SaveLoggerRunIDsOnTrainStart(output_dir))
```
mmcdermott: Should we make this callback be part of the training configs so hydra instantiate handles it, rather than doing it manually?

src/MEDS_EIC_AR/utils.py
```diff
@@ -596,20 +622,41 @@ def apply_saved_logger_run_ids(trainer_cfg: DictConfig, run_dir: Path) -> None:
 
     loggers = [e for e in raw_entries if e is not None and hasattr(e, "get")]
 
-    log_dir = Path(run_dir) / "loggers"
+    primary_log_dir = Path(run_dir) / "loggers"
```
mmcdermott: Why not just take a list of directories to check? But also, this seems like design smell -- are we just not putting some IDs in places we should elsewhere?

src/MEDS_EIC_AR/__main__.py
```diff
@@ -231,7 +239,17 @@ def generate_trajectories(cfg: DictConfig):
             "you're running this for any other purpose, set inference.do_sample=true."
         )
 
-    apply_saved_logger_run_ids(cfg.trainer, Path(cfg.model_initialization_dir))
+    # Layered restore: prefer a generation save-point under ``output_dir`` over the
+    # training save-point under ``model_initialization_dir``. On a fresh ``output_dir``
+    # this falls through to the training run id (the original behavior); on subsequent
+    # generation resumes it reads the prior generation's saved id, so the
```
mmcdermott: Why don't we want generation to always be associated with the pre-trained logger? Maybe we should just not bother saving logger run IDs during generation at all, and always read from the trainer?

src/MEDS_EIC_AR/utils.py
```diff
@@ -32,18 +32,38 @@
 logger = logging.getLogger(__name__)
 
 
+def _logger_class_candidates(logger_name: str) -> list[type]:
```
mmcdermott: Why do we need this? This seems fishy. Can you explain? Why can't we just inspect the versions of things we have installed and deterministically load the right logger class?

src/MEDS_EIC_AR/training/callbacks.py
mmcdermott: Done — dropped the historical bug context from the docstring. The callback now describes only what it does (write to <trainer.default_root_dir>/loggers/) and how that wires up; the issue link in the commit is enough for archaeology.

src/MEDS_EIC_AR/training/callbacks.py
mmcdermott: Done — moved `from MEDS_EIC_AR.utils import save_logger_run_ids` to a top-level import.

src/MEDS_EIC_AR/training/callbacks.py
mmcdermott: Good catch — done. Constructor takes no args now and reads `Path(trainer.default_root_dir)` at on_train_start. Same root every other output (checkpoints, hydra config, environment.txt) is anchored at, so a single `output_dir` knob in the trainer config controls the whole run's on-disk layout.

src/MEDS_EIC_AR/__main__.py
mmcdermott: Removed — the comment is gone along with the `trainer.callbacks.append(...)` line it was annotating (the callback is now hydra-instantiated instead, see the next comment thread).

src/MEDS_EIC_AR/__main__.py
mmcdermott: Done — added `configs/trainer/callbacks/save_logger_run_ids.yaml` (zero-arg, just the `_target_`) and wired it into the default callbacks set. Removed the manual `trainer.callbacks.append(...)` from `__main__.pretrain`. Generation's separate `callbacks/generation.yaml` deliberately doesn't pull it in (it's training-only).

src/MEDS_EIC_AR/utils.py
mmcdermott: Reworked entirely — you're right, this was a design smell. Removed the `fallback_dir` parameter and dropped the saved-id write from generation altogether; generation now `apply_saved_logger_run_ids(cfg.trainer, Path(cfg.model_initialization_dir))` and never saves its own ids. Single source of truth, no orphan write to read back. Tied to your other comment about generation being associated with the pre-trained logger.

src/MEDS_EIC_AR/__main__.py
mmcdermott: Agreed — that's exactly the new design. Generation only reads from `model_initialization_dir/loggers/`, never writes its own ids to `output_dir`. The trailing `save_logger_run_ids(...)` is gone, and `apply_saved_logger_run_ids` lost its `fallback_dir` parameter. Issue #131 is fixed by removing the orphan write rather than papering over it. The kill-and-resume test `test_generate_trajectories_attaches_to_training_id_and_does_not_save_to_output_dir` pins both halves.

src/MEDS_EIC_AR/utils.py
mmcdermott: You were right — the dance was unnecessary. Standardized the logger configs (`wandb.yaml`, `mlflow.yaml`, `csv.yaml`) to use the canonical `lightning.pytorch.loggers.*` import path (matches the rest of the codebase, which already uses `lightning.pytorch.callbacks.*` and `lightning.pytorch.trainer.Trainer`). Lightning's loggers module re-exports both classes unconditionally — the heavyweight backend imports are deferred to instantiation time inside Lightning itself — so I dropped `_logger_class_candidates` entirely and now just do `isinstance(logger, MLFlowLogger)` / `isinstance(logger, WandbLogger)` against the top-level imports.

src/MEDS_EIC_AR/utils.py
```diff
@@ -25,57 +25,27 @@
 from pathlib import Path
 
 import torch
-from lightning.pytorch.loggers import Logger
+from lightning.pytorch.loggers import Logger, MLFlowLogger, WandbLogger
```
mmcdermott, line 28: Will this crash if Wandb is not installed on the local machine?

mmcdermott: No — verified the import path. Lightning gates the actual `wandb` package import via `RequirementCache("wandb>=0.12.10")` at `lightning/pytorch/loggers/wandb.py:48`, and that gate is only checked inside `WandbLogger.__init__` (line 309: `if not _WANDB_AVAILABLE: raise ModuleNotFoundError(...)`). The class object is unconditionally exported from `lightning.pytorch.loggers` even when wandb is absent, so the top-level `from lightning.pytorch.loggers import WandbLogger` succeeds, and `isinstance(logger, WandbLogger)` is fine — it can never return True without wandb being instantiable in the first place. Same story for MLFlowLogger / mlflow.

PR #155, opened 2026-04-30
fix: saturate TIMELINE//DELTA int64 cast in format_trajectories (#154)

### LINKED ISSUE #154: format_trajectories: int64 overflow in TIMELINE//DELTA conversion crashes generation when any bin's value_mean is extreme
### Bug
`format_trajectories` in `src/MEDS_EIC_AR/generation/finalize.py` (line ~421 in 0.3.0) crashes with a polars `InvalidOperationError` when *any* TIMELINE//DELTA bin in `codes.parquet` has a `value_mean` large enough to overflow int64 after the years→microseconds conversion:

```python
.when(col("code").str.starts_with(["TIMELINE//DELTA"]))
 .then([([(col("value_mean")) * (3.1557e7)]) * (1e6)].strict_cast(Int64))
 .otherwise(0)
```

The chain is `value_mean * 3.1557e7 (s/year) * 1e6 (μs/s)`, so int64 overflows whenever `value_mean > 9.22e18 / 3.16e13 ≈ 2.92e5` years. Any time-delta bin whose stored mean is past ~292,000 years will trip this.

### Repro on real MIMIC-IV data
Running `MEICAR_generate_trajectories` on `MEDS-EIC-AR==0.3.0` against full MIMIC-IV (preprocessed via the same `MEICAR_process_data` flow with default `bin_numeric_values` config), with rolling generation (`max_new_tokens=2048`, `frac_seq_len_as_context=0.75`, `rolling_context_size=384`):

```
polars.exceptions.InvalidOperationError: conversion from `f64` to `i64` failed in column 'value_mean' for 1 out of 2429 values: [1.8985e19]
```

`1.8985e19` years isn't a meaningful clinical duration — it's a polluted bin mean, almost certainly the consequence of a single outlier numeric_value in MIMIC-IV upstream raw data getting averaged into one of the bin's row group during `fit_quantile_binning`.

The crash is **stochastic** because it only triggers when the model samples a token whose bin is the polluted one. In a 12-task-tier sweep, `tier_5/hospital_discharge` failed, while `tier_5/random_2`, `tier_5/post_admission_24h`, `tier_10/hospital_discharge`, and many others succeeded.

### Two-part fix
1. **Robust conversion** in `format_trajectories`: replace `strict_cast(Int64)` with a saturating cast (`saturating_cast` if available, or clip to `i64.MAX` first then cast). Even with cleaner bin stats, a single bad upstream value shouldn't take down a multi-hour generation run that's already produced 6 valid trajectory parquets.
2. **Robust bin reduction** during `bin_numeric_values` / `fit_quantile_binning`: compute `value_mean` as a trimmed mean or median rather than the raw mean so a single outlier numeric_value doesn't poison an entire bin's representative. (This is upstream MEDS-transforms territory, but probably reportable there too.)

### Versions
- `MEDS-EIC-AR==0.3.0`, `meds-torch-data==0.9.0`, `MEDS-transforms==0.6.7`, `polars==1.40.1`
- Python 3.12, single GPU, lightning 2.5.x

mmcdermott:
## Summary

- Fixes #154: `format_trajectories` no longer crashes on polluted `TIMELINE//DELTA` bins. The `value_mean × seconds_per_unit × 1e6 → Int64` cast is now a saturating cast (clip to the Int64-safe f64 range, then cast strict), so a single bad bin can't take down a multi-hour generation run that's already produced valid trajectory parquets.
- Adds `_warn_on_overflowing_delta_bins`, a one-shot warning that names the offending `(code, value_mean)` pairs at finalize time so users can chase the upstream pollution (`fit_quantile_binning` / `bin_numeric_values` mean reduction) rather than just seeing nonsensical timestamps.
- Widens the delta-cast arithmetic to f64 (`value_mean.cast(Float64) * seconds_per_unit * 1e6`). f32 multiplication on ~1e19 inputs is lossy enough that even clipped values promote to `2**63` on the way back to Int64; the explicit f64 cast keeps the bound representable. Side effect: non-polluted-bin timestamps shift by ≤3µs (harmless at clinical scales). The `format_trajectories` doctest is updated to match the f64-precision values.

The upstream half of the two-part fix the issue calls for — robust bin reduction via trimmed mean / median in `bin_numeric_values` — lives in MEDS-transforms and is out of scope here.

## Test plan

- [x] New `tests/test_format_trajectories_overflow.py` with three regression tests:
    - polluted-bin input no longer raises and produces output for the unaffected trailing token
    - non-polluted-bin path matches the new f64-precision timestamp exactly (locks the `value_mean → Float64 → Int64` chain so a future return to f32 arithmetic is a visible failure)
    - polluted bins surface in a `WARNING` log naming the offending bin
- [x] Verified all three fail on `main` for the right reasons (the polars `InvalidOperationError: conversion from f32 to i64 failed ... [5.9911e32]` from #154) before applying the fix
- [x] Existing `format_trajectories` doctest re-pinned to the new microsecond-precise output and passes
- [x] `tests/test_generate_trajectories.py` + all `src/MEDS_EIC_AR/generation/` doctests still pass (15/15)

## Notes for reviewers

- The "Int64-safe f64 max" constant is `2**63 - 1024` (= `9223372036854774784.0`), not the naïve `2**63 - 1` — the latter rounds *up* in f64 to `2**63`, which trips the strict cast even after "clip to Int64 max". This is the bit that took the longest to track down; the comment in `finalize.py` calls it out.
- The warning is scoped to the n_vocab code-metadata DataFrame, not the per-token exploded one, so its cost is independent of trajectory length.


src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -390,6 +435,7 @@ def format_trajectories(
     )
     code_metadata = _get_code_metadata(base_dataset)
     seconds_per_unit = _timeline_delta_seconds_per_unit(code_metadata)
+    _warn_on_overflowing_delta_bins(code_metadata, seconds_per_unit)
```
mmcdermott: This should not happen here -- instead, this should happen in the main CLI, before we do anything else. And, if there are time deltas this big, it should really just cause a crash, as we won't be able to decode trajectories to longitudinal timelines, right?

src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -340,40 +381,44 @@ def format_trajectories(
...
-        │            ┆ 18:54:31.399968       ┆                     ┆                       ┆               │
-        │ 239684     ┆ 2010-05-11            ┆ 2010-05-11 18:30:00 ┆ TIMELINE//END         ┆ null          │
-        │            ┆ 18:54:31.399968       ┆                     ┆                       ┆               │
-        └────────────┴───────────────────────┴─────────────────────┴───────────────────────┴───────────────┘
+        ┌────────────┬─────────────────┬─────────────────────┬─────────────────────────────┬───────────────┐
+        │ subject_id ┆ time            ┆ prediction_time     ┆ code                        ┆ numeric_value │
+        │ ---        ┆ ---             ┆ ---                 ┆ ---                         ┆ ---           │
+        │ i64        ┆ datetime[μs]    ┆ datetime[μs]        ┆ str                         ┆ f32           │
+        ╞════════════╪═════════════════╪═════════════════════╪═════════════════════════════╪═══════════════╡
+        │ 239684     ┆ 2010-05-11      ┆ 2010-05-11 18:00:00 ┆ TIMELINE//DELTA//years//val ┆ 0.000003      │
```
mmcdermott: Why did this change in this PR? This doesn't seem like it should have changed.

src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -420,7 +466,21 @@ def format_trajectories(
...
+                # finalize step. ``_warn_on_overflowing_delta_bins`` above logs which bin(s)
+                # tripped this so users can chase the upstream pollution.
+                #
+                # The explicit ``cast(Float64)`` matters: ``value_mean`` is Float32, and f32
+                # arithmetic on ~1e19 inputs is lossy enough that clipping in f32 still leaves
+                # a value that promotes to ``2**63`` (one ULP above Int64 max) on the way to
+                # the cast. Doing the multiply + clip in f64 keeps the bound representable.
+                (pl.col("value_mean").cast(pl.Float64) * seconds_per_unit * 1_000_000)
+                .clip(_INT64_F64_MIN, _INT64_F64_MAX)
+                .cast(pl.Int64)
```
mmcdermott: Should you put the cast outside of the when-then-otherwise anyways?

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,149 @@
...
+import pytest
+from meds import DataSchema, LabelSchema
+from meds_torchdata import MEDSPytorchDataset
+
+from MEDS_EIC_AR.generation import finalize
+from MEDS_EIC_AR.generation.finalize import TIMELINE_DELTA_TOKEN, format_trajectories
+
+# value_mean (in years) large enough that value_mean * 3.1557e7 * 1e6 overflows Int64.
+# Mirrors the polluted bin reported in issue #154 ("1.8985e19 years").
+_POLLUTED_VALUE_MEAN = 1.8985e19
```
mmcdermott: Should you construct this automatically from the max range of int64, instead of just the # reported?

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,149 @@
...
+        },
+        schema={"code_idx": pl.Int64, "code": pl.Utf8, "value_mean": pl.Float32},
+    )
+
+
+def _build_base_dataset() -> SimpleNamespace:
+    """Minimal stand-in for ``MEDSPytorchDataset``: only needs ``schema_df``."""
+    schema_df = pl.DataFrame(
+        {
+            DataSchema.subject_id_name: [42],
```
mmcdermott: you can use LabelSchema.subject_id_name as well, which will look a little cleaner here.

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,149 @@
...
+            ],
+            "value_mean": [0.000003, _POLLUTED_VALUE_MEAN, None],
+        },
+        schema={"code_idx": pl.Int64, "code": pl.Utf8, "value_mean": pl.Float32},
+    )
+
+
+def _build_base_dataset() -> SimpleNamespace:
+    """Minimal stand-in for ``MEDSPytorchDataset``: only needs ``schema_df``."""
+    schema_df = pl.DataFrame(
```
mmcdermott: col-name to vals list will be a bit cleaner.

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,149 @@
...
+                f"{TIMELINE_DELTA_TOKEN}//years//value_[POLLUTED)",
+                "DISCHARGE",
+            ],
+            "value_mean": [0.000003, _POLLUTED_VALUE_MEAN, None],
+        },
+        schema={"code_idx": pl.Int64, "code": pl.Utf8, "value_mean": pl.Float32},
+    )
+
+
+def _build_base_dataset() -> SimpleNamespace:
```
mmcdermott: Why not just mock the dataset?

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,149 @@
...
+    """Minimal stand-in for ``MEDSPytorchDataset``: only needs ``schema_df``."""
+    schema_df = pl.DataFrame(
+        {
+            DataSchema.subject_id_name: [42],
+            LabelSchema.prediction_time_name: [datetime(2024, 1, 1, 12, 0, 0)],  # noqa: DTZ001
```
mmcdermott: Why the noqa?

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Agreed — flipped to a CLI-startup validator (`validate_timeline_delta_bins_in_int64_range`) wired into `MEICAR_generate_trajectories` after the datamodule is instantiated. Saturating cast removed; format_trajectories goes back to the strict cast (with the new comment explaining the upstream guard). Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Reverted — that drift was only there because the saturating-cast path widened to f64 to keep the bound representable. With the validator approach the cast is unchanged from main and so is this expected output. Restored in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Good catch, hoisted it. Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f — the cast now wraps the whole when-then-otherwise as a single trailing `.cast(pl.Int64)`.

tests/test_format_trajectories_overflow.py
mmcdermott: Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f — the test value_mean is now `2.0 * (2**63 - 1) / (seconds_per_year * 1e6)` so it tracks the cast width / unit if either ever changes.

tests/test_format_trajectories_overflow.py
mmcdermott: The new tests target the validator (a path → exception function) directly and don't need a dataset at all — they just write a synthetic codes.parquet to tmp_path. Resolved in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

tests/test_format_trajectories_overflow.py
mmcdermott: Cleaned up in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f — the test now passes a list of dicts (col-name → val) into pl.DataFrame, no separate schema dict.

tests/test_format_trajectories_overflow.py
mmcdermott: Resolved in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f — the new tests don't construct a schema_df, so neither subject_id_name reference is needed.

tests/test_format_trajectories_overflow.py
mmcdermott: Removed in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f — same as above, the new tests don't construct datetimes.

src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -181,6 +191,54 @@ def _timeline_delta_seconds_per_unit(code_metadata: pl.DataFrame) -> float:
     return normalize_time_unit(units.pop())[1]
 
 
+def validate_timeline_delta_bins_in_int64_range(code_metadata_fp: Path) -> None:
```
mmcdermott: doctest this. Use yaml_to_disk to set up the code metadata df, or make the function take in a df instead of a path.

src/MEDS_EIC_AR/__main__.py
```diff
@@ -174,6 +178,14 @@ def generate_trajectories(cfg: DictConfig):
 
     D = instantiate(cfg.datamodule)
 
+    # Reject vocabularies with TIMELINE//DELTA bins whose ``value_mean * seconds_per_unit *
+    # 1e6`` would overflow Int64 microseconds (issue #154). Doing this before the checkpoint
+    # load / predict pass means a polluted bin fails the run in the first second instead of
+    # blowing up partway through hours of generation — and even a "saturating" decode of such
+    # a bin would yield an uninterpretable trajectory, so refusing to start is the right
+    # behavior. The fix lives upstream in the bin-reduction stage; this is the downstream guard.
+    validate_timeline_delta_bins_in_int64_range(D.config.code_metadata_fp)
```
mmcdermott: The way you're calling this removes the need for having two separate functions for code metadata, right? D has the dataset instantiated.

tests/test_format_trajectories_overflow.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def _row(code: str, idx: int, value_mean: float | None) -> dict:
+    """Encode a vocab row in (sum, n_occurrences) form so value_mean computes back to ``value_mean``."""
+    if value_mean is None:
+        return {"code": code, "code/vocab_index": idx, "values/n_occurrences": 0, "values/sum": 0.0}
+    return {"code": code, "code/vocab_index": idx, "values/n_occurrences": 1, "values/sum": value_mean}
+
+
+def test_validator_passes_on_clean_vocab(tmp_path: Path):
```
mmcdermott: Almost all of these can be doctests.

mmcdermott (commented):
None of these tests would have failed before this issue was fixed, as they just test new code you added. Can you add a test that is like the synthetic generation test (e.g., from the CLI) but with a dataset with a polluted time delta bin and verify it crashes as expected and with an ok time delta bin but a polluted non-time-delta bin and show the latter doesn't crash?

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Done — validator now takes `code_metadata: pl.DataFrame` instead of a path, and the docstring carries three doctests (clean vocab passes, polluted bin raises with the offender named, non-delta overflow is correctly ignored). Inline `pl.DataFrame` literals in the doctest avoid yaml_to_disk + parquet IO entirely. Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

src/MEDS_EIC_AR/__main__.py
mmcdermott: Yes — confirmed and collapsed. Removed `_load_code_metadata` (the path-based duplicate), promoted `_get_code_metadata` to public `get_code_metadata`, and the call site in `__main__.generate_trajectories` is now a single composition: `validate_timeline_delta_bins_in_int64_range(get_code_metadata(D.train_dataloader().dataset))`. One metadata loader, one doctestable validator. Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

tests/test_format_trajectories_overflow.py
mmcdermott: Done — replaced the standalone test file with three doctests on `validate_timeline_delta_bins_in_int64_range` itself (clean / polluted / non-delta-overflow). Deleted `tests/test_format_trajectories_overflow.py` outright. Done in 45efe7a876f0a39d341c095ebf14b9097c0f0b3f.

src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -181,6 +183,100 @@ def _timeline_delta_seconds_per_unit(code_metadata: pl.DataFrame) -> float:
...
+    int64_max = (1 << 63) - 1
+    overflow = code_metadata.filter(
+        pl.col("code").str.starts_with(TIMELINE_DELTA_TOKEN)
+        & ((pl.col("value_mean").cast(pl.Float64) * seconds_per_unit * 1_000_000).abs() > int64_max)
+    ).select("code", "value_mean")
+    if overflow.height == 0:
+        return
+    pairs = overflow.to_dicts()
+    raise ValueError(
+        f"{overflow.height} TIMELINE//DELTA bin(s) have value_mean * {seconds_per_unit} "
```
mmcdermott: Can you reword this message so the value_mean being too large is more directly stated? E.g., just say "have value_mean that, in microseconds, exceeds the maximum int64 range" or something like that. The rest is good, it is just the beginning right now that looks like a weird math equation.

src/MEDS_EIC_AR/generation/finalize.py
```diff
@@ -620,4 +723,10 @@ def finalize_predictions(
             logger.debug(f"rmdir({rank_outputs_dir}) did not succeed: {exc}")
 
 
-__all__ = ["finalize_predictions", "format_trajectories", "write_rank_output"]
+__all__ = [
```
mmcdermott: Why is this here? Do we need this?

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Reworded — done. Now opens with "have a value_mean that, when converted to microseconds (× ... s/unit × 1e6), exceeds the Int64 range". The math is still in there as a parenthetical for diagnostic value but the lead is the plain-English version. Doctest's ELLIPSIS pattern updated to match.

src/MEDS_EIC_AR/generation/finalize.py
mmcdermott: Removed entirely — `generation/__init__.py` does explicit named imports (`from .finalize import (finalize_predictions, format_trajectories, get_code_metadata, ...)`), which is the actual export gate. `__all__` only controls `from finalize import *` behavior, and the codebase doesn't use star imports anywhere. The block was just dead config.

mmcdermott:
Good catch — added `tests/test_format_trajectories_overflow_cli.py`. Two end-to-end tests that copy the preprocessed cohort to a tmp dir, mutate `metadata/codes.parquet` to introduce a polluted bin (one polluting a TIMELINE//DELTA row, one polluting an HR row), and run `MEICAR_generate_trajectories` via subprocess. The delta-polluted case asserts non-zero exit + the polluted bin code surfaces on stderr; the non-delta-polluted case asserts the CLI runs to completion. These would have failed before the validator existed (the delta case crashed deep in finalize; the non-delta case never crashed but proves the validator doesn't over-reach).

PR #164, opened 2026-09-02
fix: sample trajectories from the model's full next-token distribution (closes #159)

### LINKED ISSUE #159: Trajectory sampling is silently truncated to top-50, not calibrated draws from the model
## Summary

`MEICAR_generate_trajectories` does not sample trajectories from the model's own next-token
distribution. It samples from a **top-50 truncated** version of it, because
`Model._generate_chunk` never sets `top_k` and therefore inherits `transformers`'
`GenerationConfig` class default of `top_k=50`.

This is a silent, unintended mode-seeking bias. Generated trajectories are enriched for
high-likelihood continuations, so anything downstream that treats them as Monte Carlo draws
from the model — calibration, risk estimates, probabilistic predictions resolved from
trajectory ensembles (#11), REACH/SCOPE-style probability recording (#96, #95) — is computing
against the wrong distribution.

## Where

`src/MEDS_EIC_AR/model/model.py`, in `Model._generate_chunk`:

```python
generation_config = GenerationConfig(
    max_new_tokens=max_new_tokens,
    do_sample=do_sample,
    num_beams=1,
    temperature=1.0,
    pad_token_id=pad_id,
    bos_token_id=None,
    eos_token_id=self.HF_model.config.eos_token_id,
)
```

`num_beams=1` and `temperature=1.0` are set explicitly, which reads as "plain ancestral
sampling" — but `top_k` is absent, and `GenerationConfig().top_k` is `50`. HF adds a
`TopKLogitsWarper` whenever `top_k is not None and top_k != 0`, so every sampled trajectory is
drawn from the renormalized top-50 head of a 3954-token vocabulary.

(To be clear: this is *not* beam search. It is multinomial sampling from a truncated
distribution.)

## Evidence

Measured by generating a batch, then teacher-forcing the result back through the model and
computing the **rank** of each sampled token within that step's next-token distribution
(rank 0 = argmax). Model: 4-layer Llama, `head_dim=64`, vocab 3954, trained on a tiny cohort
derived from the public MIMIC-IV demo. Prompt length 128, 256 new tokens, batch 16.

| configuration | `top_k` | tokens | max rank | mean rank | frac. rank ≥ 50 | distinct tokens |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| repo default (`top_k` unset) | 50 | 3719 | **49** | 14.3 | **0.0000** | 475 |
| `top_k=0` (truncation off) | 0 | 3865 | 3949 | 868.7 | **0.7661** | 2035 |

The repo default never once draws a token ranked 50 or worse — a hard cap, not a soft
preference. With truncation disabled, **76.6%** of sampled tokens come from outside the top-50
set, and the mean sampled rank is 869.

Caveat on magnitude: this model is deliberately small and undertrained, so its next-token
distribution is high-entropy and the excluded mass is at its most extreme. A well-trained model
would show a lower mean rank and less excluded mass. The truncation itself is structural and
applies regardless of how well the model is trained.

Reproduce with the standalone `probe_sampling_calibration.py` posted in the first comment below;
happy to upstream it into `tests/` if useful.

## Suggested fix

Set the sampling knobs explicitly rather than inheriting HF defaults, so the intent is visible
at the call site:

```python
generation_config = GenerationConfig(
    ...,
    temperature=1.0,
    top_k=0,     # no truncation -- calibrated ancestral sampling
    top_p=1.0,
    ...
)
```

Two follow-on considerations:

1. If truncated sampling is ever wanted deliberately, it should be an explicit config knob
   under `inference:`, not an inherited default.
2. `SGLangBackend.generate_chunk` currently translates neither `top_k` nor `top_p` (its
   docstring says so and defers to #82). SGLang's own defaults happen to be non-truncating
   (`top_k=1073741824`, `top_p=1.0`), so **SGLang is already doing the calibrated thing today**
   and HF is not — the two backends do not currently sample from the same distribution. Once
   `top_k`/`top_p` become real knobs they must be plumbed through the SGLang adapter too, or
   the backends will silently disagree in the other direction.

## Observed cross-backend consequence

Over 40 sampled trajectories on the same checkpoint and cohort, via `MEICAR_generate_trajectories`:

| | HF (top-50) | SGLang (untruncated) |
| --- | ---: | ---: |
| distinct codes per trajectory (mean) | 89.5 | 246.4 |
| distinct codes generated overall | 609 | 3233 |

Jaccard similarity of the two generated code vocabularies: **0.18**.

mmcdermott:
Closes #159.

## Assessment

The report is correct, and I reproduced it against `transformers==4.57.1`:

```python
>>> GenerationConfig(max_new_tokens=4, do_sample=True, num_beams=1, temperature=1.0,
...                  pad_token_id=0, bos_token_id=None, eos_token_id=1).top_k
50
```

`Model._generate_chunk` set `do_sample`, `num_beams` and `temperature` explicitly but not `top_k`,
so HF resolved it from the class default and installed a `TopKLogitsWarper` on every call. Both the
single-chunk and the rolling path go through that one constructor, so *every* trajectory this repo
has produced under `inference.do_sample=true` was a draw from the renormalized top-50 head of the
vocabulary rather than from the model's next-token distribution.

## Fix

Pin `top_k=0` and `top_p=1.0` alongside the knobs already set. Untruncated ancestral sampling is the
only sampling mode there is, so this is stated as the sampling law at the call site rather than
exposed as a config knob — there is deliberately no way to ask for truncated sampling.

## Regression test — `tests/test_sampling_calibration.py`

Per review, the test compares generation's draws against **the model's own** next-token
distribution, which is what "calibrated" means and what can be read straight off a forward pass. No
learned generative process in between:

```python
p_model   = softmax(model.HF_model(**model._hf_inputs(batch)).logits[0, -1])   # P
p_sampled = <bincount of 10k single-token draws through Model.generate>        # P-hat
assert total_variation(p_model, p_sampled) <= 0.20
```

The forward pass is routed through `Model._hf_inputs` so it sees exactly the ids and mask
`generate` would build from the same batch. Under the sampling law this PR pins, HF applies no
logits warper at all, so that softmax *is* the distribution `generate` should be drawing from.

Two fixture choices keep the test from passing for the wrong reason:

- **Vocabulary of 512**, ten times the cutoff, so truncation removes most of the distribution rather
  than a sliver of tail.
- **The output head's weights are scaled 12x after init.** A freshly initialized model is nearly
  uniform (entropy within 1% of `log(512)`), and agreement with a uniform `P` would also hold for a
  sampler that ignored the model entirely. Sharpening puts `P` 0.50 in TV away from uniform while
  still leaving ~47% of its mass outside the top 50. That premise is asserted, not assumed.

Measured over six model seeds x two generation seeds:

| statistic | truncated to top 50 | untruncated (this PR) | threshold |
| --- | ---: | ---: | ---: |
| TV(P, P-hat) | 0.47 – 0.52 | 0.069 – 0.078 | <= 0.20 |
| distinct tokens / 10k draws | 50 (hard cap) | 472 – 489 | >= 200 |
| TV(P, uniform) *(premise)* | 0.47 – 0.50 | same | >= 0.25 |

`distinct <= 50` is a hard cap under truncation rather than a tendency: every draw conditions on the
same prompt, so the retained top-50 set is identical across draws.

Alongside the behavioral check the file carries two cheap guards:

- a direct assertion on the `GenerationConfig` handed to the backend, on **both** the single-chunk
  and rolling paths — this is what will name the problem if `top_k=0` is ever dropped, and it is
  where `temperature` is pinned, since a high-entropy `P` makes the behavioral test insensitive to
  temperature;
- a fault-injected negative control that reinstates `top_k=50` inside a wrapper backend and asserts
  the probe rejects it. Since truncation is not a supported mode there is no production knob to
  flip. This matters more than it would otherwise: with the model supplying its own ground truth, a
  probe that quietly stopped measuring has no external reference to catch it.

Verified in both directions — reverting the two-line fix fails the config test on both paths and the
behavioral test at `TV=0.466`, with the intended messages. Runs in ~1.7s; no training.

Nothing in the file asserts anything about what HF's defaults happen to be: the property under test
is our own sampling law, and the negative control keeps the probe honest without reaching for a
third-party default to do it.

## Follow-on for the SGLang branch

`feat/sglang-backend` carries a comment in `SGLangBackend.generate_chunk` saying `top_p`/`top_k` are
untranslated because "none of today's callers set them (see `Model._generate_chunk` — the only call
site)". That is stale once this merges. The behavior is still correct — SGLang's own defaults are
non-truncating, so it was already doing the calibrated thing and now the two backends agree — but the
comment needs updating on merge, and any future translation of these fields must keep both backends
on the same law.

## Testing

- `uv run pytest --ignore=tests/grammar`: 78 passed, 4 skipped.
- `pre-commit run`: clean.
- The grammar suite is left to CI. It is unaffected in principle — its vocabulary is 20 tokens, so
  the top-50 cutoff was never binding there.


src/MEDS_EIC_AR/model/model.py
```diff
@@ -1002,6 +1002,14 @@ def _generate_chunk(
         explicitly. ``eos_token_id`` is read from ``self.HF_model.config.eos_token_id`` rather than
         passed in, since both callers always want the model's configured EOS.
 
+        **Sampling law.** Untruncated ancestral sampling is the only sampling mode this model
+        supports: when ``do_sample`` is true, generation draws from the model's full next-token
```
mmcdermott: This says "it only supports one mode", then says "when do_sample is true", which implies do_sample could ever be false. Which is it?

tests/calibration/_markov.py
mmcdermott, line 1: Is there an easier test than this? If top_k filters the set of logits that could be seen to only the most likely 50, can we actually just take a model over a vocabulary > 50, observe the next token logits, then check on a single sampling step that the selected tokens by the generation loop are calibrated? E.g., the model (randomly initialized) on a prompt has observed distribution (pulled from the model's own forward) over next tokens of P, the empirical distribution after many calls through generate is \hat{P}, assert that P = \hat{P}? Then we don't need any of this other machinery, and it tests directly what we care about, without configuring things in the intermediate state, right?

src/MEDS_EIC_AR/model/model.py
mmcdermott: Fair — that read as self-contradictory. The sentence was trying to say "there is only one *sampling* mode" while `do_sample` legitimately also selects greedy, which is a decoding mode but not a sampling one. Rewritten to separate the two:

> **Sampling law.** `do_sample` picks between the two decoding modes there are: `False` greedy-decodes, `True` samples. There is exactly one sampling mode, and it is untruncated ancestral sampling — `temperature=1.0`, `top_k=0`, `top_p=1.0`, `num_beams=1` — so a sampled token is a draw from the model's full next-token distribution and nothing narrower. Truncated sampling is not offered. Every one of those four fields is set explicitly, including the ones that merely restate a HF default, because the failure mode being guarded against is an inherited default silently changing the distribution being sampled.

On the follow-up question of whether greedy should exist at all: it currently underpins the 100% grammar-validity assertion in `tests/grammar/test_cli.py`, two in-process grammar tests, ~10 doctests in this file documenting the rolling loop's chunk/budget/EOS semantics, and deterministic assertions in `tests/test_generation_backend.py`. Filed #166 to make the correctness test stop depending on it, which is the prerequisite for revisiting the parameter itself. Out of scope here.

tests/calibration/_markov.py
mmcdermott: You're right, and I've rewritten it your way. `tests/calibration/` is gone; it's now a single file, `tests/test_sampling_calibration.py`, with no training and no Markov chain.

Your framing is the better one for a reason I'd underweighted: what makes a sample *calibrated* is agreement with **the model**, and the model's distribution is right there to be read off a forward pass. The Markov chain was measuring "the model learned the chain" *and* "sampling is calibrated" together, and only the second is under test here. Removing it also removes the whole class of flakiness where a training run that converged badly fails a sampling test.

The probe is now exactly as you describe:

```python
p_model  = softmax(model.HF_model(**model._hf_inputs(batch)).logits[0, -1])   # P
p_sampled = <bincount of 10k single-token draws through Model.generate>        # P-hat
assert total_variation(p_model, p_sampled) <= 0.20
```

Routed through `Model._hf_inputs` so the forward pass sees exactly the ids and mask `generate` would build from the same batch. With the sampling law this PR pins, HF applies no logits warper at all, so that softmax *is* the distribution `generate` should be drawing from — nothing intermediate to configure.

Two deliberate choices on the fixture, both to stop the test passing for the wrong reason:

- **Vocab 512** (10x the cutoff), so truncation removes most of the distribution rather than a sliver of tail.
- **The output head's weights are scaled by 12x after init.** A freshly initialized model's next-token distribution is nearly uniform — entropy within 1% of `log(512)` — and against a uniform `P`, `P-hat ≈ P` would also hold for a sampler that ignored the model entirely and drew uniformly. Sharpening gives a `P` that is 0.50 in TV away from uniform while still keeping ~47% of its mass outside the top 50. The test asserts that "far from uniform" premise explicitly, so if someone retunes the fixture into vacuity it fails rather than passes.

Measured over six model seeds x two generation seeds:

| statistic | truncated to top 50 | untruncated | threshold |
| --- | ---: | ---: | ---: |
| TV(P, P-hat) | 0.47 – 0.52 | 0.069 – 0.078 | <= 0.20 |
| distinct tokens / 10k draws | 50 (hard cap) | 472 – 489 | >= 200 |
| TV(P, uniform) *(premise)* | 0.47 – 0.50 | same | >= 0.25 |

Tighter than the old fixture on every axis, and it runs in ~1.7s instead of ~5s of training. Kept from the old version: the direct `GenerationConfig` assertion (both generation paths), the guard that HF still defaults `top_k=50`, and the fault-injected negative control. That last one now matters more than before — with the model supplying its own ground truth, a probe that silently stopped measuring would have no external reference to catch it, so the control re-runs the identical probe against a backend that reinstates `top_k=50` and asserts it's rejected.

One thing the new probe deliberately gives up: near-maximal entropy makes it insensitive to a `temperature` bug. That's covered declaratively instead — `test_generate_chunk_requests_untruncated_sampling` asserts `temperature == 1.0` alongside `top_k`/`top_p`/`num_beams`, so the whole sampling law is pinned even where the behavioral test can't see it.

Verified in both directions: reverting the two-line fix fails the config test (both paths) and the behavioral test with `TV=0.466`.

tests/test_sampling_calibration.py
```diff
@@ -0,0 +1,310 @@
...
+            "any value that is neither None nor 0 (its class default is 50), which silently turns "
+            "sampling into a draw from the truncated top-k head instead of from the model's own "
+            "next-token distribution."
+        )
+        assert cfg.top_p == 1.0, f"GenerationConfig.top_p is {cfg.top_p!r}, not 1.0 (no nucleus truncation)."
+        assert cfg.temperature == 1.0, f"GenerationConfig.temperature is {cfg.temperature!r}, not 1.0."
+        assert cfg.num_beams == 1, f"GenerationConfig.num_beams is {cfg.num_beams!r}, not 1."
+
+
+def test_generation_config_default_top_k_is_still_the_hazard():
```
mmcdermott: I don't think you need this test. It is testing a property that we want to not need to care about from HF. The behavior test really ought to be the only one we need (though the one that checks our parameters are set as desired is still worth it, to be clear).

mmcdermott: Agreed — removed in 26f782e, along with the now-unused import and the module docstring's reference to it.

You're right that it was reaching for a third-party default to establish something we don't depend on. Nothing else in the file consults HF's defaults: the code sets `top_k=0` outright, the config test checks that we do, and the behavioral test compares draws against the model's own softmax. So the only thing that assertion could ever have done is break CI on a transformers upgrade that changed a value we never read.

The thing it was nominally protecting — "is the probe still sensitive to truncation?" — is already covered directly by `test_probe_detects_top_k_truncation`, which reinstates truncation itself rather than inferring sensitivity from an upstream default. That's the stronger guard of the two anyway.

Down to three tests: the config assertion (kept, as you say), the behavioral check, and the negative control.

PR #167, opened 2026-09-02
fix: record ARM64 distributions in uv.lock (closes #161)

### LINKED ISSUE #161: uv sync fails on aarch64: uv.lock's `av` entry records no wheels or sdist
## Summary

`uv sync` fails on aarch64 (ARM) Linux hosts because `uv.lock`'s `av` entry records no
distributions at all — no wheels, no sdist.

```
error: Distribution `av==17.0.1 @ registry+https://pypi.org/simple` can't be installed because
it doesn't have a source distribution or wheel for the current platform
```

Reproduced on a DGX Spark (aarch64, Python 3.11) with `uv 0.11.3`, running
`uv sync --extra sglang --frozen`.

## Cause

`av` is pulled in by `sglang`, but only under an ARM-only marker:

```toml
{ name = "av", marker = "(platform_machine == 'aarch64' and sys_platform == 'linux') or ..." }
```

and its package stanza in `uv.lock` is empty:

```toml
[[package]]
name = "av"
version = "17.0.1"
source = { registry = "https://pypi.org/simple" }
```

No `wheels = [...]`, no `sdist = ...`. The lock was resolved in an environment where that marker
evaluated false (x86_64), so uv never recorded any distribution for it — and on ARM, where the
marker *is* true, there is nothing to install.

This is not a missing-wheel problem upstream: `av-17.0.1-cp311-abi3-manylinux_2_28_aarch64.whl`
exists on PyPI and installs fine.

## Workaround

`uv pip install -e .` resolves and installs `av` normally.

## Suggested fix

Re-lock with aarch64 included in the resolution environments, e.g. in `pyproject.toml`:

```toml
[tool.uv]
environments = [
    "sys_platform == 'linux' and platform_machine == 'x86_64'",
    "sys_platform == 'linux' and platform_machine == 'aarch64'",
    "sys_platform == 'darwin'",
]
```

so the lock carries distributions for the ARM-only branch too. Worth doing given ARM boxes
(DGX Spark, GH200/GB200, Apple silicon CI) are increasingly where this gets run.

mmcdermott:
Closes #161. Branches off `feat/sglang-backend` (#117), where the `sglang` extra and the affected
lock entry live.

## Assessment

Confirmed. On `feat/sglang-backend`, the `av` stanza in `uv.lock` is four lines with nothing to
install:

```toml
[[package]]
name = "av"
version = "17.0.1"
source = { registry = "https://pypi.org/simple" }
```

No `sdist`, no `wheels`. `av` is reachable only under an aarch64-only marker via the `sglang`
extra; the lock is resolved on x86_64 where that marker is false, so uv never explored the branch
and recorded no distribution for it. On ARM, where the marker holds, there is nothing there. As the
issue says, this is not an upstream gap — the aarch64 wheels exist on PyPI.

## Fix

```toml
[tool.uv]
required-environments = ["sys_platform == 'linux' and platform_machine == 'aarch64'"]
```

**`required-environments`, not `environments`.** The issue suggested `environments`, which does fix
the symptom, but it also *restricts* resolution to the enumerated set — anything not listed
(Windows, most obviously) silently loses support. `required-environments` only widens what the lock
must cover, which is the actual requirement here.

**Linux ARM only.** I tried adding `sys_platform == 'darwin' and platform_machine == 'arm64'` for
the Apple-silicon case the issue mentions, and backed it out: sglang publishes no macOS ARM wheels
at 0.5.9, so requiring coverage there forks the resolution back to 0.5.2 and drops
`mlx-lm`/`mlx-metal`. That is 250 lines of lock churn plus a downgrade of the very pin this
branch's Blackwell install recipe is written against — to cover a platform the `sglang` extra can't
run on anyway. The reasoning is recorded in the `pyproject.toml` comment so it isn't re-litigated.

## Effect on the lock

22 insertions, 12 deletions. `av` gains its sdist and four aarch64 wheels
(`cp311-abi3` and `cp314-cp314t`, manylinux and musllinux). The rest is a handful of intra-package
dependency markers simplified as a consequence of the widened resolution space
(`mlx-lm`, `nvidia-cudnn-cu12`, `nvidia-cufft-cu12`, `nvidia-cusolver-cu12`, `nvidia-cusparse-cu12`).
**No package version changes.**

## Testing

- `uv lock --check`: passes.
- `uv sync --dry-run` on x86_64: unchanged apart from the local project re-link, so no regression on
  the platform CI runs.
- `pre-commit run`: clean.

**Needs confirmation on ARM hardware**, which I don't have: `uv sync --extra sglang --frozen` on the
DGX Spark. The defect itself is visible in the file and is fixed there, but the end-to-end install
is worth a look before this is trusted. Note also that this lock was regenerated with uv 0.8.13
while the report came from 0.11.3; the lock revision is unchanged (3).

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,11 +99,13 @@ Note also that `sglang_demo.yaml` sets `mem_fraction_static: 0.85`, a fraction o
 Spark shares one 128 GB pool between CPU and GPU, so that reserves ~109 GB and leaves the host close to
 the OOM killer; `0.2`–`0.4` is ample for small models.
 
-> [!NOTE]
-> `uv sync --frozen` currently fails on aarch64 hosts: `uv.lock`'s `av` entry records no wheel or sdist,
-> because it was resolved on x86_64 where `av`'s aarch64-only marker is false. Use `uv pip install -e .`
-> until the lockfile is regenerated with an aarch64 resolution environment. See
-> [#161](https://github.com/mmcdermott/MEDS_EIC_AR/issues/161).
+`uv sync --frozen` works on aarch64 hosts because `pyproject.toml` declares
+`tool.uv.required-environments` with the Linux ARM branch, which forces `uv.lock` to carry
+distributions for it. **Keep that setting if you re-lock.** Without it, a dependency reachable only
+under an ARM marker — `av`, pulled in by the `sglang` extra — gets a lock entry with no wheel and no
+sdist, since the resolver never explores that branch on an x86_64 host. `uv sync` on ARM then fails
+with "can't be installed because it doesn't have a source distribution or wheel for the current
+platform".
 
 ## Usage
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,3 +139,21 @@ select = [
 
 [tool.ruff.lint.pep8-naming]
 extend-ignore-names = ["*MEDS*", "F", "L", "M", "D", "LR_*", "*GPT*"]  # Ignore MEDS in variable names
+
+[tool.uv]
+# Force the lock to carry distributions for ARM64 as well as the x86_64 host it is usually
+# resolved on. Without this, a dependency reachable *only* under an ARM marker gets a package
+# stanza with no `sdist` and no `wheels` at all: the resolver never explored that branch, so it
+# recorded nothing to install, and `uv sync` on an ARM host fails with "can't be installed because
+# it doesn't have a source distribution or wheel for the current platform". `av`, pulled in by the
+# `sglang` extra under an aarch64-only marker, hit exactly that on a DGX Spark.
+#
+# `required-environments` rather than `environments`: the latter would *restrict* resolution to the
+# listed set and silently drop support for anything not enumerated (Windows, for one). This only
+# widens what the lock must cover.
+#
+# Only the Linux ARM branch is listed. Adding macOS ARM was tried and rejected: `sglang` publishes
+# no macOS ARM wheels at 0.5.9, so requiring coverage there forks the resolution back to 0.5.2 and
+# drops `mlx-lm`/`mlx-metal` — a 250-line lock churn and a downgrade of the pin this branch's
+# install recipe is written against, to cover a platform the `sglang` extra cannot run on anyway.
+required-environments = ["sys_platform == 'linux' and platform_machine == 'aarch64'"]
--- a/uv.lock
+++ b/uv.lock
@@ -7,6 +7,9 @@ resolution-markers = [
     "python_full_version < '3.12' and sys_platform == 'linux'",
     "python_full_version < '3.12' and sys_platform != 'linux'",
 ]
+required-markers = [
+    "platform_machine == 'aarch64' and sys_platform == 'linux'",
+]
 
 [[package]]
 name = "aiohappyeyeballs"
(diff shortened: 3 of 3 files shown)


mmcdermott:
Tested on real hardware (DGX Spark, aarch64, Python 3.11, `uv 0.11.3`). Branch confirmed not stale: `origin/feat/sglang-backend..origin/fix/uv-lock-aarch64-161` is 0 commits.

**Verified fixed.** In a clean worktree off this branch:

```
$ uv sync --extra sglang --frozen
...
EXIT=0
```

That is the exact command that previously failed with ``Distribution `av==17.0.1` can't be installed because it doesn't have a source distribution or wheel for the current platform``. The resulting venv has `av 17.0.1`, `sglang 0.5.9`, `torch 2.9.1`, `transformers 4.57.1`, and the `av` lock entry now carries the aarch64 wheels plus an sdist.

`required-environments` rather than `environments` is the right call, and the comment explaining why (the latter would *restrict* rather than widen, silently dropping Windows) is worth keeping. Rejecting macOS ARM on the grounds that `sglang` publishes no wheels there at 0.5.9 also matches what I see.

One expectation to set, not a defect in this PR: the resulting `torch` is the CPU build (`torch.version.cuda is None`). That is the separate, already-documented aarch64 behaviour — CUDA torch still needs the `uv pip install -e .` + explicit-wheel route from the README's SM 12.1 section. So `uv sync --frozen` now *succeeds* on ARM, but it is not by itself a GPU-ready install. Might be worth one sentence in the README saying so, since the note this PR replaces was the thing previously steering ARM users to `uv pip install -e .`.

Note this branch and #169 both rewrite the same `mem_fraction_static` paragraph in `README.md`; whichever lands second will want a trivial conflict resolution.

mmcdermott:
Follow-up, since "does a naive install get CPU torch?" came up and deserves a precise answer. Short version: **x86_64 is unaffected, and this PR is not the cause on ARM either.**

**This PR does not touch torch.** The `torch` stanza in `uv.lock` is byte-identical before and after — same 24 wheels, same dependency markers. The whole diff is: add `av`'s ARM distributions, plus marker *simplifications* on `mlx-lm` and a few `nvidia-*` entries that fall out of declaring a second required environment. No resolution change for torch on any platform.

**On x86_64 you get CUDA torch, as always.** That's upstream PyPI metadata for `torch==2.9.1`:

```
nvidia-cuda-nvrtc-cu12==12.8.93; platform_system == "Linux" and platform_machine == "x86_64"
nvidia-cudnn-cu12==9.10.2.21;    platform_system == "Linux" and platform_machine == "x86_64"
... 16 CUDA deps, all gated identically
```

so an x86_64 Linux install pulls the full nvidia stack and is CUDA-enabled. Nothing here changes that, with or without the `sglang` extra.

**On aarch64 you get CPU torch — but that is PyTorch's packaging, not this repo's.** PyTorch publishes no CUDA ARM wheel to PyPI at 2.9.1; the ARM wheel reports `2.9.1+cpu`. Confirmed:

- `uv sync --frozen` with **no extras** on ARM → `torch 2.9.1+cpu`, `version.cuda = None`, `nvidia-cublas-cu12` not installed.
- The lock has **zero** references to `download.pytorch.org`; the wheel comes straight from `files.pythonhosted.org`.
- The same x86_64-gated markers are present on `origin/dev`, so this predates the SGLang work entirely.

So the ordering is worth being clear about: before this PR, `uv sync` on ARM **failed outright** — you got no environment at all. After it, you get a working environment whose torch happens to be CPU. That is strictly an improvement, not a regression.

**A better remedy than the README's manual index URL.** `uv`'s `--torch-backend=auto` detects the driver and picks the right index. Verified on the DGX Spark:

```
$ uv pip install --torch-backend=auto --reinstall-package torch torch==2.9.1
 + torch==2.9.1+cu130
 + triton==3.5.1
 ...
-> torch 2.9.1+cu130 | version.cuda = 13.0 | is_available = True
```

Two caveats before anyone wires this in:

1. `--torch-backend` is only on `uv pip install` in uv 0.11.3 (`UV_TORCH_BACKEND` env var); **`uv sync` does not accept it**, so it can't be folded into the sync path yet.
2. `auto` selected **cu130** here, whereas the README's SM 12.1 recipe currently specifies cu129 torch. cu130 would in principle line up better with `sgl-kernel==0.3.21+cu130`, but I have **not** run the SGLang stack on `torch 2.9.1+cu130` — my validated combination is cu129 torch + `+cu130` kernel. Worth testing before changing the documented recipe.

None of this blocks the PR. It only argues for replacing the sentence I suggested earlier with something concrete: that `uv sync` on ARM now succeeds but yields CPU torch, and `--torch-backend=auto` (or the explicit index) is the follow-up step for a GPU-ready ARM install.

PR #168, opened 2026-09-02
fix: apply SGLang's context ceiling visibly instead of letting it clamp (closes #162)

### LINKED ISSUE #162: SGLang backend silently returns 2 fewer tokens at the context-window boundary
## Summary

`SGLangBackend` silently returns **2 fewer tokens** than requested whenever
`prompt_len + max_new_tokens` reaches the model's context window — which is exactly what
`Model.generate`'s single-chunk path always requests. So every single-chunk SGLang generation is
2 tokens shorter than the equivalent HF one, with no warning.

## Evidence

Measured by calling `HFBackend.generate_chunk` and `SGLangBackend.generate_chunk` directly with
an identical prompt (`prompt_len=128`) and identical `GenerationConfig`, against a Llama with
`max_position_embeddings=512`:

| requested `max_new_tokens` | prompt + budget | HF returns | SGLang returns | deficit |
| ---: | ---: | ---: | ---: | ---: |
| 32 | 160 | 32 | 32 | 0 |
| 256 | 384 | 256 | 256 | 0 |
| 380 | 508 | 380 | 380 | 0 |
| 382 | 510 | 382 | 382 | 0 |
| 383 | 511 | 383 | **382** | 1 |
| 384 | 512 | 384 | **382** | 2 |

SGLang's usable ceiling is `max_position_embeddings - 2` total positions. HF allows the full
`max_position_embeddings`.

## Cause

`Model.generate`'s single-chunk path requests exactly the boundary:

```python
max_new_tokens=self.max_seq_len - batch.code.shape[1]
```

i.e. `prompt_len + max_new_tokens == max_seq_len`. SGLang's tokenizer-manager validator rejects
`input + max_new >= context_len`, so `SGLangBackend.__init__` sets `allow_auto_truncate=True` to
avoid a hard `ValueError`. That converts the rejection into a **silent clamp**.

The backend docstring currently describes this as matching HF's semantics:

> Setting it here matches HF's semantics by silently clamping — otherwise every first-chunk call
> in the rolling loop would raise `ValueError("Requested token count exceeds...")` at the boundary.

That's the right call to keep the run alive, but it isn't parity: it's short by 2, and the
docstring reads as though it were exact.

## Why it matters

- Cross-backend trajectory comparisons are systematically off-by-2 in length.
- In the **rolling** path the deficit compounds: each chunk that saturates the window gives back
  2 fewer tokens than the loop's budget accounting assumes, so the number of chunk iterations
  needed for a given `rolling_generation.max_new_tokens` differs between backends.

## Suggested fix

Either:

1. Have `SGLangBackend` cap the requested `max_new_tokens` at `context_len - prompt_len - 2`
   itself and **log at info/warning** when it does, so the truncation is visible; or
2. Have the shared `_generate_chunk` request `max_seq_len - prompt_len - 2` for all backends, so
   HF and SGLang agree by construction (costs HF 2 tokens, buys exact parity); or
3. At minimum, correct the docstring to state the 2-token deficit explicitly.

Observed with `sglang==0.5.9` / `sgl-kernel==0.3.21+cu130` on a DGX Spark (GB10, SM 12.1).

mmcdermott:
Closes #162. Branches off `feat/sglang-backend` (#117).

## Assessment

The report is right, and the docstring it quotes was making a claim the code doesn't support. I'd
add one thing the issue doesn't cover, which changes what the right fix is.

**The rolling path doesn't just lose two tokens — it can hang.** `_rolling_generate` advances by
`new_tokens.size(1)` and loops `while n_generated < max_new_tokens`. Its default
`rolling_context_size` is `max_seq_len - 1`, so once the window saturates every chunk has
`prompt_len = max_seq_len - 1` and `chunk_budget = 1` — two positions past SGLang's ceiling, with
only one to give back. A clamp that lands on zero returns an empty chunk, the loop makes no
progress, and it spins forever. Silent shortening is the benign case.

## Which of the three suggested fixes

Option 1 (cap in the backend, log it), not option 2. Costing HF two tokens on every chunk to match
an SGLang validator quirk buys symmetry at the price of correctness on the backend that was already
right. And capping *before* the call is what makes the zero-budget case detectable at all — the
engine clamping after the fact can only hand back an empty chunk.

## What this does

`SGLangBackend` reads `max_position_embeddings` from the exported HF directory's `config.json` — the
same number SGLang derives its context length from — and subtracts `_SGLANG_CONTEXT_RESERVE = 2`
before the call:

| case | behavior |
| --- | --- |
| request fits under the ceiling | passed through untouched |
| request over the ceiling | cut to the ceiling, warned **once per backend instance** |
| prompt leaves no room | `ValueError` naming the knob to change |
| `config.json` unreadable | budget untouched; falls back to the engine's auto-truncate, as before |

Log-once matters because the rolling loop calls `generate_chunk` once per chunk with the same
saturating shape — warning every time would bury the run's real output under hundreds of identical
lines.

The ceiling is computed against the **longest** prompt in the batch, since `max_new_tokens` is
shared across the call and rolling chunks are ragged once post-EOS padding is stripped.

`allow_auto_truncate` stays on, but demoted from mechanism to safety net: it now only catches a
residual mismatch between the reserve we subtract and the one an installed SGLang enforces, as a
shortfall rather than a mid-run crash. The comments and test docstrings that described it as
matching HF are corrected — per the issue's option 3, which is folded in rather than chosen instead.

## Behavior change worth flagging

**Rolling generation on SGLang now needs `rolling_generation.rolling_context_size <= max_seq_len - 3`**
and raises with that message otherwise. That is a new hard failure where there was previously a hang
or a silently corrupted trajectory, so I think it's the right trade — but it's a real change, not
just a diagnostic.

The nicer end state is for `GenerationBackend` to declare its context reserve and for
`_rolling_generate` to size `ctx_size` around it, so the default config just works on every backend.
That touches the protocol and the shared loop on `main`, so I've left it out of this PR; happy to
file it as a follow-up if you want it.

## Testing

- Seven new tests against the existing fake engine: pass-through below and at the ceiling, the clamp
  and its warning, log-once, the no-room raise, longest-prompt sizing, and the unreadable-config
  fallback. Plus doctests on the config reader.
- `uv run pytest --ignore=tests/grammar`: 110 passed, 4 skipped.
- `pre-commit run`: clean.

**Not verified on real hardware — this machine has no GPU.** Worth confirming on the DGX Spark that
a rolling run with `rolling_context_size = max_seq_len - 3` completes, and that the default one
raises the intended error rather than hanging.

The constant itself no longer needs re-measuring — see below.

---

## Follow-up: traced through the sglang source

Rather than leave `_SGLANG_CONTEXT_RESERVE = 2` as a measured constant, I read `sglang` v0.5.9. The
2 is exact and derivable, and my original attribution was wrong — pushed as fa02e7e (comments only,
no behavior change).

It is **not** the tokenizer-manager validator, and **not** `allow_auto_truncate`. That layer checks
`len(input) + max_new_tokens >= context_len` and, under auto-truncate, clamps to
`context_len - len(input)` — a no-op at exactly the boundary. (Its `num_reserved_tokens` addend is 0
unless Eagle speculative decoding is on.) The file even flags the split itself:
`# FIXME: unify the length validation logic with the one in the scheduler.`

The tokens go in the **scheduler**:

```python
# tp_worker.py
self.max_req_len = min(self.model_config.context_len - 1, self.model_runner.max_token_pool_size - 1)

# scheduler.py, init_req_max_new_tokens
req.sampling_params.max_new_tokens = min(
    req.sampling_params.max_new_tokens if ... is not None else 1 << 30,
    self.max_req_len - len(req.origin_input_ids) - 1,
)
```

`(context_len - 1) - len(input) - 1` = **`context_len - len(input) - 2`**, matching the reported
table exactly. That `min` is unconditional: no warning, not gated on `allow_auto_truncate`. So
`allow_auto_truncate` only decides whether a boundary request is *rejected*; it never had anything
to do with the two missing tokens.

Two things this changes:

**2 is a floor, not a guarantee.** `max_req_len` takes the `min` with `max_token_pool_size - 1`. When
the KV pool binds instead of the context length — a large model, or a small `mem_fraction_static` —
the shortfall is larger than 2. Capping by the constant keeps the common case honest; the residual
is what `allow_auto_truncate` is left on to absorb. Worth knowing alongside the
`mem_fraction_static` change in #169.

**The empty-chunk hang is confirmed from source, not inferred.** At `len(input) == context_len - 1` —
exactly what the rolling loop's default `rolling_context_size` produces once the window saturates —
the scheduler's cap evaluates to `-1`. `Req.check_finished` stops on
`len(output_ids) >= max_new_tokens`, which holds at zero output tokens, so the request completes
having generated nothing. `n_generated += 0`, and the loop spins. That is the case this PR raises on.

### DIFF
--- a/src/MEDS_EIC_AR/model/backends/sglang.py
+++ b/src/MEDS_EIC_AR/model/backends/sglang.py
@@ -44,14 +44,14 @@ Gotchas accounted for:
 from __future__ import annotations
 
 import atexit
+import json
 import logging
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import torch
 
 if TYPE_CHECKING:
-    from pathlib import Path
-
     from transformers import GenerationConfig
 
 logger = logging.getLogger(__name__)
@@ -79,6 +79,73 @@ _HF_ONLY_KWARGS: frozenset[str] = frozenset(
 #: We probe for both so a version bump in either direction doesn't silently produce empty rows;
 #: if neither key is found we raise loudly (see ``generate_chunk``).
 _SGLANG_OUTPUT_KEYS: tuple[str, ...] = ("output_ids", "token_ids")
+
+#: Positions SGLang's tokenizer-manager validator holds back from the model's context window.
+#: The validator rejects a request outright once ``input_len + max_new_tokens`` reaches the
+#: context length, and the usable ceiling measured against ``sglang==0.5.9`` is
+#: ``context_len - 2`` total positions, not ``context_len - 1``: with a 512-position model and a
+#: 128-token prompt, 382 new tokens is the largest request that comes back at full length, while
+#: 383 and 384 both come back as 382. HF's ``generate``, by contrast, allows the full
+#: ``context_len``. So an SGLang chunk is inherently up to two tokens shorter than the HF chunk
+#: the shared rolling loop asks for, and this backend subtracts that itself rather than letting
+#: the engine clamp silently. The constant is empirical and version-sensitive; if a future SGLang
+#: changes the validator, this is the number to re-measure.
+_SGLANG_CONTEXT_RESERVE = 2
+
+
+def _read_context_length(hf_model_dir: Path | str) -> int | None:
+    """Read ``max_position_embeddings`` from an HF model directory's ``config.json``.
+
+    This is the number SGLang derives its own context length from when it loads the directory, so
+    reading it here lets the adapter apply SGLang's ceiling before the call instead of discovering
+    it as a silent clamp afterwards.
+
+    Returns ``None`` rather than raising if the file is missing, unreadable, or has no
+    ``max_position_embeddings`` — a backend that cannot determine the ceiling falls back to the
+    engine's own ``allow_auto_truncate`` behavior, which is what it did before. Losing the
+    proactive cap is worth strictly less than refusing to construct the backend at all.
+
+    Examples:
+        >>> import json, tempfile
+        >>> from pathlib import Path
+        >>> with tempfile.TemporaryDirectory() as d:
+        ...     _ = (Path(d) / "config.json").write_text(json.dumps({"max_position_embeddings": 512}))
+        ...     _read_context_length(d)
+        512
+
+        Missing or malformed configs degrade to ``None``:
+
+        >>> with tempfile.TemporaryDirectory() as d:
+        ...     print(_read_context_length(d))
+        None
+        >>> with tempfile.TemporaryDirectory() as d:
+        ...     _ = (Path(d) / "config.json").write_text("{not json")
+        ...     print(_read_context_length(d))
+        None
+        >>> with tempfile.TemporaryDirectory() as d:
+        ...     _ = (Path(d) / "config.json").write_text(json.dumps({"vocab_size": 40}))
+        ...     print(_read_context_length(d))
+        None
+    """
+    config_fp = Path(hf_model_dir) / "config.json"
+    try:
+        config = json.loads(config_fp.read_text())
+    except (OSError, ValueError) as e:
+        logger.warning(
+            f"Could not read {config_fp} to determine the model's context length "
+            f"({type(e).__name__}: {e}). SGLang's own auto-truncation will apply instead, which "
+            "clamps silently."
+        )
+        return None
+
+    context_len = config.get("max_position_embeddings")
+    if not isinstance(context_len, int):
+        logger.warning(
+            f"{config_fp} has no integer ``max_position_embeddings`` (got {context_len!r}); "
+            "SGLang's own auto-truncation will apply instead, which clamps silently."
+        )
+        return None
+    return context_len
 
 
 def _strip_padding_to_lists(input_ids: torch.Tensor, attention_mask: torch.Tensor | None) -> list[list[int]]:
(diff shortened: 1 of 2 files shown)


mmcdermott:
Tested on real GPU hardware (DGX Spark, GB10 / SM 12.1; `sglang==0.5.9`, `sgl-kernel==0.3.21+cu130`, `torch==2.9.1+cu129`). Branch confirmed not stale: `origin/feat/sglang-backend..origin/fix/sglang-context-boundary-162` is 0 commits.

**The diagnosis is right and the arithmetic checks out.** Independent measurement against a live engine reproduces the ceiling exactly, and the clamp lands on the right number:

| requested | prompt+budget | HF | SGLang |
| ---: | ---: | ---: | ---: |
| 382 | 510 | 382 | 382 |
| 383 | 511 | 383 | 382 |
| 384 | 512 | 384 | 382 |

With the PR applied, a 384-token request on a 128-token prompt returns exactly 382, an in-budget 382 request is passed through untouched, and both `allowed == 0` and `allowed < 0` raise with the `rolling_context_size` remedy named. 11 of 12 behavioural checks pass.

Two problems, though — one of which I think blocks merge.

### 1. The warning is suppressed in practice, so the shortfall is still invisible

This is the PR's stated purpose ("apply the ceiling *visibly*"), and on real hardware it does not happen. Constructing `sgl.Engine(...)` raises the **root logger level to ERROR**:

```
BEFORE engine boot: module.effective=30 root.level=30 isEnabledFor(WARNING)=True
AFTER  engine boot: module.effective=40 root.level=40 isEnabledFor(WARNING)=False
```

`SGLangBackend._cap_max_new_tokens` emits its warning *after* the engine exists, so it is dropped. Measured end-to-end: the clamp fires correctly (382 tokens returned) and **zero** clamp warnings are observable. A canary `logger.warning` on the same module logger immediately after boot is likewise swallowed. Same mechanism silences the repo's own `Writing trajectory N to ...` INFO lines during SGLang runs.

`ServerArgs.log_level` defaults to `'error'` and SGLang applies it process-wide, so this is not specific to the CLI.

The unit tests don't catch it because they pair a `_FakeEngine` (no real logging reconfiguration) with `caplog.at_level("WARNING")`, which force-sets the level. Mock-green, real-silent.

Suggestions: capture and restore the root level around `sgl.Engine(...)`, or pass an explicit `log_level` into the engine kwargs, or emit through a handler the backend owns. Whatever the fix, it'd be worth one test that asserts visibility *after* a real (or realistically-reconfiguring) engine init.

### 2. Default rolling generation now raises instead of running

`_rolling_generate` defaults `rolling_context_size` to `max_seq_len - 1`, so a saturated window gives `allowed = context_len - (max_seq_len - 1) - 2 = -1` and the new `ValueError` fires. The repo's own gated test fails on this branch:

```
ValueError: SGLang cannot generate any new tokens for a 14-token prompt against a
16-position context window ... Set rolling_generation.rolling_context_size to at most 13
```

and so does the CLI on a MIMIC-demo-derived cohort at `max_seq_len=512` (`rolling_generation.max_new_tokens=600`, everything else default) — 0 trajectory files written. Passing `rolling_generation.rolling_context_size=509` makes that same run complete and write output, so the remedy works; it just isn't the default.

Raising is better than the empty-chunk spin the docstring describes, so the direction is right — but as it stands every rolling SGLang run needs a hand-computed magic number, and the shipped test config doesn't supply one.

Suggestion: let the backend advertise its reserve and have `_rolling_generate` subtract it, e.g. an optional `context_reserve: int = 0` on the `GenerationBackend` protocol with

```python
reserve = getattr(self._backend, "context_reserve", 0)
ctx_size = min(rolling_context_size or (self.max_seq_len - 1), self.max_seq_len - 1 - reserve)
```

That keeps HF untouched and makes the default work for SGLang without user configuration.

### Not caused by this PR, but found while testing it

With the ceiling respected (`rolling_context_size=13`), the gated grammar test still fails — on **correctness**, not the ceiling: 128/128 held-out samples grammar-invalid. I reproduced the identical failure on this PR's base with the same settings, so it is pre-existing and neither caused nor fixed here. Filed separately.

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -80,6 +80,89 @@
...
+#: **Two is a floor, not a guarantee.** When the KV pool is the binding term instead — a large
+#: model, or a small ``mem_fraction_static`` — ``max_req_len`` is smaller than ``context_len - 1``
+#: and the shortfall grows. Capping by this constant keeps the common case honest; the residual is
+#: what ``allow_auto_truncate`` is left on to absorb.
+#:
+#: The subtraction also goes negative: at ``len(input) == context_len - 1`` the scheduler's cap is
+#: ``-1``, and ``Req.check_finished`` stops on ``len(output_ids) >= max_new_tokens``, which holds
+#: at zero output tokens. The request completes having generated nothing — which is why
+#: :meth:`SGLangBackend._cap_max_new_tokens` raises rather than letting the rolling loop receive an
+#: empty chunk it would spin on.
```
mmcdermott, line 108: Why do you have `#:` for these comments, not `#`? Just curious.

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -80,6 +80,89 @@
...
+_SGLANG_CONTEXT_RESERVE = 2
```
mmcdermott, line 109: Per the comments on one of the other PRs, can this be read from some engine property somewhere? Or inferred via some other property? If not, then this is fine, but if it can be read directly, that'd be better.

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -80,6 +80,89 @@
...
+
+
+def _read_context_length(hf_model_dir: Path | str) -> int | None:
```
mmcdermott, line 112: If you want this to just get the max position embeddings, can you call it that? Also, why is this better than using the max sequence length?

mmcdermott, line 112: Also, could this function just use the actual config itself, or the functions that read the actual config, rather than doing manual JSON parsing? Wouldn't that save some code and/or be more robust?

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -204,21 +287,74 @@ def __init__(
...
+        fits. Otherwise returns the ceiling and logs the shortfall once.
+
+        Raises:
+            ValueError: If the prompt is so long that no new tokens fit under the ceiling. This is
+                reachable from the rolling loop, whose default ``rolling_context_size`` of
+                ``max_seq_len - 1`` leaves only one position for new tokens — one fewer than the
+                reserve needs. Raising beats returning an empty chunk, which the rolling loop
+                would read as "no progress" and spin on forever.
+        """
+        if self._context_len is None:
```
mmcdermott, line 326: Why is this clause right?

src/MEDS_EIC_AR/model/backends/sglang.py
mmcdermott: `#:` is Sphinx/autodoc syntax. A comment block starting `#:` immediately above a module-level assignment becomes that constant's documentation in generated API docs, the way a docstring does for a function; a plain `#` is invisible to the doc build. It also renders in some editors on hover.

So it is meaningful rather than decorative, but it is only worth it on module-level constants that are part of the module's vocabulary — which is where this file uses it. Plain `#` for ordinary inline commentary is right.

src/MEDS_EIC_AR/model/backends/sglang.py
mmcdermott: Yes, it can, and it now is — in #172, stacked on this branch.

`sgl.Engine` exposes `scheduler_info`, which carries the number the scheduler actually enforces:

```
{'status': 'ready', 'max_total_num_tokens': 34276968, 'max_req_input_len': 10}
```

Both ceilings descend from `max_req_len = min(context_len - 1, kv_pool - 1)`, and `max_req_input_len = max_req_len - 5`, so the reported value recovers both. That let me delete both `_SGLANG_CONTEXT_RESERVE` and the input-side constant, leaving one documented gap constant.

Beyond removing magic numbers, it fixes a case the arithmetic here cannot express: when the **KV pool** binds rather than the context length, the true ceiling is smaller and depends on runtime memory. Verified the reported value against measurement on two models (506 for a 512-position model, 10 for a 16-position one) and both match what I observed the engine enforcing.

src/MEDS_EIC_AR/model/backends/sglang.py
mmcdermott: Both fair.

On naming: it did just read `max_position_embeddings`, so the name was doing the reader no favours. Moot as of #172 — the ceilings no longer come from the config at all, so `_read_context_length` and the `_context_len` attribute are gone. The only config field still read is `head_dim`, and the helper is now named for what it does (`_read_hf_config_value(dir, key)`).

On why not `max_seq_len`: they are the same number — `export_lightning_to_hf_dir` writes `max_position_embeddings` from `Model.max_seq_len`. The backend read the config only because it has no reference to the `Model` that owns `max_seq_len`; it is handed a directory. So it was rediscovering a value it had effectively already been given, which is a fair criticism of the old shape. Reading it from the engine instead removes the duplication rather than papering over it.

mmcdermott: Agreed, and changed in #172. It now goes through `transformers.AutoConfig.from_pretrained(...)` instead of `json.loads`.

Concretely more robust for our one remaining field: `head_dim` is often absent from `config.json` and derived from `hidden_size / num_attention_heads`. Hand-parsing sees `None` and silently skips the pre-flight check; `AutoConfig` resolves it the way the model itself does, so the check fires on models where it previously would not have. It also drops the bespoke `_config_int` type-guard helper.

Worth noting it did surface one thing: configs without `model_type` cannot be loaded by `AutoConfig`. Real exports always write it, but the unit-test fixtures were writing minimal configs without it, so the test helper now defaults it — which makes those fixtures closer to what the export actually produces.

src/MEDS_EIC_AR/model/backends/sglang.py
mmcdermott: It is not, really — good catch. `if self._context_len is None: return requested` means "if we could not read the config, silently skip the cap and let `allow_auto_truncate` do whatever it does". Given the whole point is to stop relying on that silent behaviour, deferring to it exactly when we are least informed is backwards.

Resolved in #172 rather than defended: the ceiling comes from the engine, so there is no unknown-context branch left. If the engine does not report a usable value, construction raises instead of proceeding on a guess — a ceiling guessed too high means silently truncated prompts with nothing downstream able to tell.

PR #169, opened 2026-09-02
fix: make SGLang's head-dim floor and memory reservation safe by default (closes #163)

### LINKED ISSUE #163: SGLang on Blackwell SM 12.1 (DGX Spark): working recipe, FlashInfer head-dim limit, and a misleading gated-test default
## Summary

SGLang now runs end-to-end on Blackwell SM 12.1 (GB10 / DGX Spark), which #117 could not
achieve. It needs one install step that the `sglang` extra can't express, plus one model-shape
constraint. Filing so the recipe is written down and so the gated test isn't misleading.
Relates to #88, #117, #122.

## 1. `sgl-kernel` from PyPI has no SM 12.1 kernels

PyPI `sgl-kernel` wheels carry kernel images only up to `sm_120a`, and `sm_120a` is **not**
forward-compatible with SM 12.1. The first RMSNorm dies:

```
RuntimeError: RMSNorm failed with error code no kernel image is available for execution on the device
```

The `+cu130` builds from SGLang's own index do contain `sm_121a`:

```bash
uv pip install --index-url https://docs.sglang.ai/whl/cu130/ "sgl-kernel==0.3.21+cu130"
```

`0.3.21` is exactly what `sglang==0.5.9` pins, and a `+cu130` kernel build works against a
`torch==2.9.1+cu129` runtime — the binding ABI that matters is the torch *version*, not the CUDA
variant. Verify coverage before trusting a wheel:

```bash
cuobjdump --list-elf .venv/lib/python3.11/site-packages/sgl_kernel/sm100/common_ops.abi3.so \
  | grep -o 'sm_[0-9]*[a-z]*' | sort -u
# +cu130 : sm_100 sm_110a sm_120a sm_121a sm_90     <- works
# PyPI   : sm_100a sm_101a sm_120a sm_80 sm_87 ...  <- no sm_121a
```

Two further runtime requirements on this hardware:

- `TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas` — the CUDA 12.8 `ptxas` bundled with torch 2.9.1
  doesn't know `sm_121a`.
- **`.venv/bin` must be on `PATH`.** SGLang JIT-compiles kernels at engine startup and shells out
  to `ninja`, which is installed as a venv console script. Without it the engine dies with
  `FileNotFoundError: [Errno 2] No such file or directory: 'ninja'` inside `tvm_ffi`.

## 2. FlashInfer rejects small head dims — and our gated test uses one

With `attention_head_dim=32`, SGLang's default FlashInfer attention backend aborts the scheduler
subprocess:

```
RuntimeError: Error in function 'BatchPrefillWithRaggedKVCacheDispatched' ...
FlashInfer Internal Error: Invalid configuration : NUM_MMA_Q=2 NUM_MMA_D_QK=2 NUM_MMA_D_VO=2
NUM_MMA_KV=4 NUM_WARPS_Q=4 NUM_WARPS_KV=1
```

which surfaces in the parent as `SIGQUIT` / exit `-9` — i.e. it looks like an OOM, not a
configuration error. `attention_head_dim=64` works with the default backend;
`engine_kwargs.attention_backend=triton` works at 32.

`tests/grammar/conftest.py` defaults `--grammar-model-head-dim` to **32**, so
`tests/grammar/test_cli_sglang.py` fails out of the box on any host where it actually runs. It
passes the shape constraint when invoked as:

```bash
pytest tests/grammar/test_cli_sglang.py --grammar-model-head-dim=64
```

Suggested: either bump the grammar fixture default to 64, or have `sglang_demo.yaml` set
`attention_backend: triton`, or raise a clear error from `SGLangBackend` when
`head_dim < 64` and the FlashInfer backend is selected.

## 3. `sglang_demo.yaml`'s `mem_fraction_static: 0.85` is rough on unified memory

DGX Spark has 128 GB shared between CPU and GPU. `mem_fraction_static: 0.85` reserves ~109 GB and
leaves the host with a couple of GB, which invites the OOM killer if anything else is resident.
`0.2`–`0.4` was ample for a small model. Worth a comment in the config noting the value is a
fraction of *total* memory, and that unified-memory hosts want a much lower number.

## Status

With the above, `MEICAR_generate_trajectories backend=sglang_demo` completes end-to-end and
writes valid trajectories, validated against a tiny cohort built from the public MIMIC-IV demo.

mmcdermott:
Closes #163. **Stacked on #168** (issue #162), which introduces the config reader this generalizes —
review that one first.

The install recipe from part 1 of the issue already landed in 6202ced. This is the code and config
half: parts 2 and 3, plus the gated test's model shape.

## FlashInfer's head-dim floor

`SGLangBackend` now reads `head_dim` from the exported HF config and refuses to start when it is
below 64 **and** no `attention_backend` was named, with an error giving both remedies.

The reason this deserves a hard error rather than a doc note: the failure it prevents is an abort
inside the scheduler subprocess that reaches the parent as SIGQUIT / exit `-9`. A shape constraint
that presents as an out-of-memory kill sends you looking at memory, which is exactly where the time
goes.

Naming a backend explicitly is honored — that is both the documented remedy and what keeps this
check from wrongly blocking a future FlashInfer that lifts the floor. The check only fires when
*we* are the ones letting SGLang pick.

## The gated test's model shape

I went with the issue's second option rather than its first. `sglang_demo.yaml` now sets
`attention_backend: triton`, so `tests/grammar/test_cli_sglang.py` runs against the grammar
fixture's own model shape.

Raising `--grammar-model-head-dim` to 64 would have doubled the width of a model that every grammar
test pretrains on CPU — the fixture is already the dominant cost in `run_tests_ubuntu` — to
accommodate a backend that one gated GPU test uses. Triton has no floor and costs the CPU path
nothing. The knob keeps its 32 default and now carries that reasoning in its help text, so the next
person reading it doesn't re-file the same issue.

## mem_fraction_static

`sglang_demo.yaml` drops 0.85 → **0.3**; `sglang.yaml` keeps 0.85 with the caveat written down. Both
configs and the README now say explicitly that it is a fraction of *total* device memory reserved up
front, and that unified-memory hosts want a much lower number. 0.3 is ample for the models the demo
config targets and is safe on a dedicated GPU too.

## Also here

The config reader from #168 is generalized from context-length-only to "parse `config.json` once,
pull typed fields". Both users degrade to skipping their check rather than refusing to construct the
backend — these are pre-flight diagnostics, and none is worth failing a run over. The README's
FlashInfer bullet is updated to say the backend now catches this, and its `+backend.engine_kwargs...`
override loses the stray `+` (the key exists in both configs now).

## Testing

- Four new tests: refusal below the floor, acceptance when a backend is named, acceptance exactly at
  the floor, and no-check when `head_dim` is absent. Plus doctests on the reader and the guard.
- `uv run pytest --ignore=tests/grammar`: 116 passed, 4 skipped.
- `pre-commit run`: clean.

**Not verified on real hardware — no GPU here.** Worth confirming on the DGX Spark:

1. That `tests/grammar/test_cli_sglang.py` now passes with no `--grammar-model-head-dim` override,
   i.e. that `attention_backend: triton` really does carry `head_dim=32` end to end. This is the
   claim I'm least able to check, since it rests on triton having no floor.
2. That `mem_fraction_static: 0.3` is enough for the demo run there.
3. That 64 is the right floor for the installed `sglang==0.5.9` — it's from the issue's measurements
   (32 aborts, 64 works), and the constant records that provenance so it can be re-measured.

### DIFF
--- a/src/MEDS_EIC_AR/model/backends/sglang.py
+++ b/src/MEDS_EIC_AR/model/backends/sglang.py
@@ -93,59 +93,127 @@ _SGLANG_OUTPUT_KEYS: tuple[str, ...] = ("output_ids", "token_ids")
 _SGLANG_CONTEXT_RESERVE = 2
 
 
-def _read_context_length(hf_model_dir: Path | str) -> int | None:
-    """Read ``max_position_embeddings`` from an HF model directory's ``config.json``.
+#: Smallest attention head dimension SGLang's default FlashInfer attention backend can dispatch.
+#: Below this it aborts the scheduler subprocess with ``FlashInfer Internal Error: Invalid
+#: configuration``, which reaches the parent as SIGQUIT / exit ``-9`` — indistinguishable at a
+#: glance from an out-of-memory kill, and so a genuinely expensive thing to debug. Measured against
+#: ``sglang==0.5.9``: ``head_dim=32`` aborts, ``head_dim=64`` works. The ``triton`` attention
+#: backend has no such floor.
+_FLASHINFER_MIN_HEAD_DIM = 64
 
-    This is the number SGLang derives its own context length from when it loads the directory, so
-    reading it here lets the adapter apply SGLang's ceiling before the call instead of discovering
-    it as a silent clamp afterwards.
 
-    Returns ``None`` rather than raising if the file is missing, unreadable, or has no
-    ``max_position_embeddings`` — a backend that cannot determine the ceiling falls back to the
-    engine's own ``allow_auto_truncate`` behavior, which is what it did before. Losing the
-    proactive cap is worth strictly less than refusing to construct the backend at all.
+def _read_hf_config(hf_model_dir: Path | str) -> dict:
+    """Parse an HF model directory's ``config.json``, or return ``{}`` if it can't be read.
+
+    Degrading to an empty dict rather than raising is deliberate: every use of this config is a
+    pre-flight check that improves diagnostics, and none of them is worth refusing to construct
+    the backend over. A missing config costs the checks, not the run.
 
     Examples:
         >>> import json, tempfile
         >>> from pathlib import Path
         >>> with tempfile.TemporaryDirectory() as d:
-        ...     _ = (Path(d) / "config.json").write_text(json.dumps({"max_position_embeddings": 512}))
-        ...     _read_context_length(d)
-        512
+        ...     _ = (Path(d) / "config.json").write_text(json.dumps({"head_dim": 64}))
+        ...     _read_hf_config(d)
+        {'head_dim': 64}
 
-        Missing or malformed configs degrade to ``None``:
+        Missing and malformed configs both degrade to ``{}``:
 
         >>> with tempfile.TemporaryDirectory() as d:
-        ...     print(_read_context_length(d))
-        None
+        ...     _read_hf_config(d)
+        {}
         >>> with tempfile.TemporaryDirectory() as d:
         ...     _ = (Path(d) / "config.json").write_text("{not json")
-        ...     print(_read_context_length(d))
-        None
-        >>> with tempfile.TemporaryDirectory() as d:
-        ...     _ = (Path(d) / "config.json").write_text(json.dumps({"vocab_size": 40}))
-        ...     print(_read_context_length(d))
-        None
+        ...     _read_hf_config(d)
+        {}
     """
     config_fp = Path(hf_model_dir) / "config.json"
     try:
         config = json.loads(config_fp.read_text())
     except (OSError, ValueError) as e:
         logger.warning(
-            f"Could not read {config_fp} to determine the model's context length "
-            f"({type(e).__name__}: {e}). SGLang's own auto-truncation will apply instead, which "
-            "clamps silently."
+            f"Could not read {config_fp} ({type(e).__name__}: {e}). The SGLang backend's "
+            "pre-flight checks on context length and attention head dimension will be skipped."
         )
-        return None
+        return {}
+    return config if isinstance(config, dict) else {}
 
-    context_len = config.get("max_position_embeddings")
-    if not isinstance(context_len, int):
-        logger.warning(
-            f"{config_fp} has no integer ``max_position_embeddings`` (got {context_len!r}); "
-            "SGLang's own auto-truncation will apply instead, which clamps silently."
-        )
+
+def _config_int(config: dict, key: str) -> int | None:
+    """Pull an integer field out of a parsed HF config, or ``None`` if it isn't there.
+
+    Examples:
+        >>> _config_int({"head_dim": 64}, "head_dim")
+        64
+        >>> print(_config_int({}, "head_dim"))
+        None
+        >>> print(_config_int({"head_dim": None}, "head_dim"))
+        None
+        >>> print(_config_int({"head_dim": "64"}, "head_dim"))
+        None
+
+        ``bool`` is a subclass of ``int`` but never a valid answer here:
+
+        >>> print(_config_int({"head_dim": True}, "head_dim"))
+        None
+    """
+    value = config.get(key)
+    if isinstance(value, bool) or not isinstance(value, int):
         return None
-    return context_len
+    return value
+
+
+def _check_head_dim_against_attention_backend(head_dim: int | None, engine_kwargs: dict) -> None:
+    """Refuse to start SGLang on a default attention backend that cannot dispatch this head dim.
+
+    SGLang defaults to FlashInfer, which aborts on ``head_dim`` below
+    :data:`_FLASHINFER_MIN_HEAD_DIM`. The abort kills the scheduler subprocess and surfaces in the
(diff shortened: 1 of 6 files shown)


mmcdermott:
Tested on real GPU hardware (DGX Spark, GB10 / SM 12.1; `sglang==0.5.9`, `sgl-kernel==0.3.21+cu130`, `torch==2.9.1+cu129`). Branch confirmed not stale: `origin/fix/sglang-context-boundary-162..origin/fix/sglang-head-dim-and-mem-163` is 0 commits.

**This one does what it says.** Every claim I could check holds up:

| check | result |
| --- | --- |
| `head_dim=32` + default backend refused at construction (before any engine boot) | PASS — `ValueError` naming both remedies |
| `head_dim=64` + default backend allowed | PASS |
| `head_dim=64` + explicit `null` allowed | PASS |
| `head_dim=32` + explicit `triton` allowed | PASS |
| unknown `head_dim` not blocked | PASS |
| `head_dim=32` + explicit `null` still refused (the `sglang.yaml` shape) | PASS |

The guard runs before `sgl.Engine(...)`, so the refusal costs nothing and — the real win — replaces a scheduler abort that surfaced as exit `-9`. I originally chased that as an OOM for a while, so this is a genuine save.

End-to-end on the CLI:

- `backend=sglang_demo` against a `head_dim=32` checkpoint now runs **out of the box** and writes trajectories. Before, it needed a hand-added `+backend.engine_kwargs.attention_backend=triton`. `attention_backend='triton'` is confirmed in the resolved `ServerArgs`.
- `backend=sglang` (production) against a `head_dim=64` checkpoint also runs and writes trajectories; the explicit `attention_backend: null` resolves to `'flashinfer'` as intended, so keeping the key present for override-without-`+` doesn't break the engine call. (I ran that one with `mem_fraction_static=0.4`; I did not test the shipped `0.85` on this unified-memory box.)

Keeping `--grammar-model-head-dim` at 32 and pointing `sglang_demo.yaml` at triton is the right trade — it avoids widening every CPU grammar pretrain for one gated GPU test, and the reasoning is now written down in the option's help text.

The `mem_fraction_static` documentation is accurate to what I measured: at `0.85` on this 128 GB unified-memory host SGLang reserved ~109 GB and left the host ~3 GB, which is how I hit the OOM killer earlier in this session. `0.3` is comfortable.

One small note: this branch and #167 both rewrite the same `mem_fraction_static` paragraph in `README.md`, so whichever lands second will likely want a trivial conflict resolution.

Caveat on scope: `_FLASHINFER_MIN_HEAD_DIM = 64` is calibrated against `sglang==0.5.9` on SM 12.1. The guard correctly no-ops when a backend is named explicitly, so a future FlashInfer that lifts the floor can be opted into without a code change — that design choice is worth keeping.

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -109,59 +109,127 @@
...
+
+        >>> _check_head_dim_against_attention_backend(32, {})
+        Traceback (most recent call last):
+            ...
+        ValueError: SGLang's default FlashInfer attention backend cannot dispatch
+        attention_head_dim=32 ...
+    """
+    if head_dim is None or head_dim >= _FLASHINFER_MIN_HEAD_DIM:
+        return
+    if engine_kwargs.get("attention_backend") is not None:
```
mmcdermott, line 222: I think you want to ensure that attention_backend can only be set to triton or flashinfer if you do something like this -- what if the user sets it to flashinfer or to a garbage string -- right now, your system will not always reject if it is a garbage string and may reject inappropriately in some cases. Unless I'm misunderstanding?

mmcdermott: You are not misunderstanding — that is a real hole, and it fails in both directions you describe. Fixed in #172, which sits on top of this branch.

Two concrete gaps in the guard as written:

1. **Explicit `flashinfer` below the floor was let through.** The check was `if engine_kwargs.get("attention_backend") is not None: return`, so naming FlashInfer deliberately skipped the very check that exists for it — and it then aborts the scheduler subprocess with exit `-9` anyway. Naming a backend does not make it dispatch a narrow head dim. The floor now applies whenever FlashInfer is what will actually run, implicit or explicit, and the error says which.
2. **A garbage string was never validated.** Now checked against SGLang's own `ATTENTION_BACKEND_CHOICES` (importable from `sglang.srt.server_args`), so a typo fails here with the valid list rather than later inside the engine.

The one case worth preserving is the one you implicitly single out: a *non-FlashInfer* backend below the floor. The floor is FlashInfer's, not a general one, so `triton` at `head_dim=32` still constructs — that is what makes `sglang_demo.yaml` work with the grammar fixture. So the rule is now "which backend will run?" rather than "did the user name one?".

New tests cover explicit-flashinfer-refused, unknown-name-refused, and non-flashinfer-below-floor-allowed, alongside the existing cases.

Everything else in this PR I verified on hardware and it holds up; this was the only issue I found.

PR #172, opened 2026-09-02
fix: stop SGLang silently truncating the prompt (closes #171)

### LINKED ISSUE #171: SGLang rolling generation produces grammar-invalid trajectories where HF does not
## Summary

Driven through the rolling-generation loop, the SGLang backend produces **grammar-invalid** trajectories where the HF backend produces 100%-valid ones from the same checkpoint. `tests/grammar/test_cli_sglang.py` fails its correctness assertion on **128/128** held-out samples.

This is **not** caused by #168/#169 — I reproduced it identically on their base (`feat/sglang-backend` @ `5649550`). Filing separately so it isn't mistaken for fallout from those PRs.

## Evidence

Hardware: DGX Spark (GB10, SM 12.1), `sglang==0.5.9`, `sgl-kernel==0.3.21+cu130`, `torch==2.9.1+cu129`.

To get past the context ceiling (#162) I ran the grammar fixture with `rolling_generation.rolling_context_size=13` (`max_seq_len=16`). Both backends in one pytest session, same fixture, same checkpoint:

```
tests/grammar/test_cli.py::test_grammar_cli_greedy_output_is_fully_grammar_valid   PASSED
tests/grammar/test_cli_sglang.py::test_sglang_greedy_output_is_fully_grammar_valid FAILED
```

So the `rolling_context_size=13` setting is not itself the cause — HF is fine with it.

The SGLang failures are uniform:

```
traj=0 subject=273: 5/15 valid (first invalid at generated position 5):
  prompt_tokens=[2, 3, 4, 5, 1],
  generated=[2, 3, 4, 5, 1, 5, 1, 2, 2, 2, 3, 3, 4, 4, 4]
traj=0 subject=279: 5/15 valid (first invalid at generated position 5):
  prompt_tokens=[9, 10, 11, 12, 13, 14, 15, 16, 17, 1],
  generated=[2, 3, 4, 5, 1, 5, 1, 2, 2, 2, 3, 3, 4, 4, 4]
traj=0 subject=287: 5/15 valid (first invalid at generated position 5):
  prompt_tokens=[6, 7, 8, 1],
  generated=[2, 3, 4, 5, 1, 5, 1, 2, 2, 2, 3, 3, 4, 4, 4]
```

Every subject and every one of the 8 trajectories emits the *same* token stream despite different prompts, goes invalid at position 5, and degenerates into runs of repeated tokens.

On the PR base the same run produces the same shape of failure (tail differs slightly: `[2,3,4,5,1,5,1,1,2,2,2,3,3,3,4]`), 128/128 invalid, first invalid at position 5.

## What it is *not*

`SGLangBackend.generate_chunk` is faithful to HF. Measured directly against a live engine on a 4-layer Llama (`head_dim=64`, vocab 3954), greedy, EOS disabled:

- **Uniform unpadded prompts, k=12 new tokens:** `SGLang(k) == HF(k)` on 3/3 rows, exactly.
- **One token at a time, appending as the rolling loop does:** `SGLang(1)×12 == HF(12)` on 3/3 rows, exactly. So the small per-chunk budgets that #168's ceiling forces are not the problem.
- **Left-padded, variable-length batch (real lengths 4/5/10/7) with an attention mask** — the shape the CLI always produces: `SGLang == HF` on 4/4 rows, exactly. So `_strip_padding_to_lists` and the mask handling are not the problem.

Nor is it fp16-vs-fp32 drift in any obvious way: those exact-match results are HF fp32 against SGLang's downcast fp16.

## Where I stopped

I did not isolate the root cause. Given per-chunk parity is exact on every shape I could construct, the divergence appears to come from the rolling loop's *interaction* with the backend rather than from a single `generate_chunk` call — the accumulation, re-priming, or per-row termination bookkeeping across chunks. That is where I would look next.

Worth noting the grammar fixture is an extreme regime (`max_seq_len=16`, ~18-token vocabulary, 400-epoch CPU pretrain), so a first step might be reproducing at a larger `max_seq_len` to see whether the effect persists or is specific to very short windows.

## Why it matters

Rolling generation is the documented path for trajectories longer than the context window (README §3.1, and the whole point of #82). Until this is understood, SGLang-backed trajectories from that path shouldn't be treated as interchangeable with HF's — the single-chunk path looks fine, but the rolling path does not.

mmcdermott:
Closes #171.

## The bug

SGLang reads at most `context_len - 7` prompt tokens. Past that it **does not reject the request** — it drops the tail of the prompt and generates from the remaining prefix. The returned tensor has the right shape and the right number of new tokens, so nothing at the API surface indicates that the model was conditioned on something the caller never asked for.

From `sglang==0.5.9`:

```
max_req_len       = min(context_len - 1, kv_pool - 1)     # managers/tp_worker.py:288
max_req_input_len = max_req_len - 5                       # managers/tp_worker.py:291
if len(origin_input_ids) >= max_req_input_len:            # managers/utils.py:111
    truncate (allow_auto_truncate) else reject
```

The comparison is `>=`, so the longest prompt read in full is `context_len - 7`. Confirmed against a live engine on a 16-position model: a 9-token prompt is honored, a 10-token prompt is not. With `allow_auto_truncate=False` the engine says so outright:

```
ValueError: Input length (11 tokens) exceeds the maximum allowed length (10 tokens).
```

Two things kept it silent:

1. `SGLangBackend` sets `allow_auto_truncate=True` unconditionally, which converts that rejection into a truncation.
2. SGLang *does* log a warning when it truncates — through a logger its own engine init has already set to ERROR (`ServerArgs.log_level` defaults to `'error'` and is applied process-wide). So the warning never appears.

`Model._rolling_generate` defaults its window to `max_seq_len - 1`, which is far above the ceiling, so **every** rolling chunk lost its prompt tail.

## Impact

On the grammar fixture (16-position model, window 15 vs ceiling 9), same checkpoint, greedy:

| backend | output | grammar-invalid |
| --- | --- | ---: |
| HF | `[2,3,4,5,1, 2,3,4,5,1, 2,3,4,5,1]` | 0 / 128 |
| SGLang (before) | `[2,3,4,5,1, 5,1, 2,2,2,3,3,4,4,4]` | **128 / 128** |

The degenerate tail is the model being re-fed a truncated prefix each chunk.

## The fix

Backends may advertise an optional `max_prompt_len`. `SGLangBackend` computes it from the model's context length; the HF backend advertises nothing and is unaffected.

- `Model._rolling_generate` clamps its window to that ceiling, so the default configuration is correct with no user intervention.
- `SGLangBackend.generate_chunk` refuses an over-long prompt with an error naming the ceiling and the remedy, rather than letting the engine quietly discard part of it. Silent corruption becomes a loud failure for any caller that builds its own prompt window.

## What this is *not*

Not a greedy-vs-sampling difference. Greedy decoding was correct throughout on prompts that fit. Per-chunk parity with the HF backend is **exact** below the ceiling — measured for uniform prompts, one-token-at-a-time decoding, and left-padded variable-length batches. Not the attention backend either: `triton` and `torch_native` fail identically, and `head_dim=32` vs `64` makes no difference.

## Relationship to #168

Same base quantity, different derived limit — adjacent but distinct:

| | formula | effect |
| --- | --- | --- |
| #168 (output) | `min(requested, max_req_len - len(input) - 1)` = `context_len - len(input) - 2` | 2 fewer tokens emitted; output still correct |
| this (input) | `max_req_input_len = max_req_len - 5`, checked with `>=` | prompt silently altered; output *wrong* |

Both descend from `max_req_len = min(context_len - 1, kv_pool - 1)`, which is why the constants look related. #168's analysis was correct but stopped at the output side.

Worth noting this also makes #168's `allowed <= 0` error unreachable from the rolling loop: once the window is clamped to `context_len - 7`, the output budget is always `>= 5`. That was the failure that made the gated test error out on that branch.

## Verification

DGX Spark (GB10, SM 12.1), `sglang==0.5.9`, `sgl-kernel==0.3.21+cu130`, `torch==2.9.1+cu129`:

- `tests/grammar/test_cli_sglang.py` — **passes** with default settings (was failing). SGLang's output is byte-identical to HF's, 0/128 grammar-invalid.
- `tests/test_sglang_backend.py` — 37 passed, including 5 new tests covering the ceiling, the boundary, the unknown-context case, and the longest-prompt-in-batch case.
- Rest of the suite (`tests/`, `src/` doctests, excluding `tests/grammar`) — 126 passed, 4 skipped.

Branched off `fix/sglang-head-dim-and-mem-163` (top of the #167/#168/#169 stack), since it builds on #168's context-ceiling work.

https://claude.ai/code/session_01WuRZgDxXbTXbZCtFo2hHdy

### DIFF
--- a/src/MEDS_EIC_AR/model/backends/sglang.py
+++ b/src/MEDS_EIC_AR/model/backends/sglang.py
@@ -109,30 +109,22 @@ _SGLANG_OUTPUT_KEYS: tuple[str, ...] = ("output_ids", "token_ids")
 _SGLANG_CONTEXT_RESERVE = 2
 
 
-#: Positions SGLang holds back from the *input* side of the context window.
+#: Fallback for SGLang's input ceiling when the engine doesn't report one.
 #:
-#: Distinct from :data:`_SGLANG_CONTEXT_RESERVE`, which bounds how many tokens SGLang will
-#: *emit*. This one bounds how many prompt tokens it will actually *read*, and exceeding it is
-#: far more damaging: SGLang does not refuse the request, it silently drops the tail of the
-#: prompt and generates from the truncated prefix, so the model is conditioned on something the
-#: caller never asked for. Derivation, from ``sglang==0.5.9``::
+#: SGLang bounds how many prompt tokens it will *read*, separately from how many it will emit
+#: (:data:`_SGLANG_CONTEXT_RESERVE`). Exceeding it is silent and destructive: the engine keeps
+#: the *oldest* tokens, discards the tail, and generates from that stale prefix, so the model is
+#: conditioned on something the caller never sent. From ``sglang==0.5.9``::
 #:
-#:     max_req_len       = min(context_len - 1, max_token_pool_size - 1)   # tp_worker.py
-#:     max_req_input_len = max_req_len - 5                                 # tp_worker.py
-#:     if len(input_ids) >= max_req_input_len: truncate or reject          # managers/utils.py
+#:     max_req_len       = min(context_len - 1, kv_pool_size - 1)   # managers/tp_worker.py
+#:     max_req_input_len = max_req_len - 5                          # managers/tp_worker.py
+#:     if len(input_ids) >= max_req_input_len: truncate             # managers/utils.py
 #:
-#: The comparison is ``>=``, so the longest prompt SGLang reads in full is
-#: ``max_req_input_len - 1``, i.e. ``context_len - 7`` whenever the context length is the
-#: binding term. Measured against a 16-position model: a 9-token prompt is honored, a 10-token
-#: prompt is already truncated.
-#:
-#: Why this is not merely SGLang's business: ``allow_auto_truncate=True`` (set unconditionally
-#: below) is what turns the rejection into a silent truncation, and SGLang's own truncation
-#: warning is emitted through a logger its engine init has already set to ERROR, so nothing
-#: surfaces. Left unchecked, ``Model._rolling_generate``'s default ``rolling_context_size`` of
-#: ``max_seq_len - 1`` sits far above this ceiling and every rolling chunk loses its prompt
-#: tail — which is exactly how SGLang came to emit grammar-invalid trajectories where the HF
-#: backend emitted valid ones (issue #171).
+#: The comparison is ``>=``, so the longest prompt read in full is ``max_req_input_len - 1``,
+#: i.e. ``context_len - 7`` when the context length is the binding term. Prefer the engine's own
+#: ``scheduler_info["max_req_input_len"]`` (see :meth:`SGLangBackend._resolve_max_prompt_len`),
+#: which also covers the case where the KV pool binds instead of the context; this constant is
+#: only the fallback for engines that don't report it.
 _SGLANG_INPUT_RESERVE = 7
 
 
@@ -409,22 +401,39 @@ class SGLangBackend:
         self._logged_context_clamp = False
         self._logged_prompt_ceiling = False
         self._engine = sgl_module.Engine(model_path=str(hf_model_dir), **self._engine_kwargs)
+        self._max_prompt_len = self._resolve_max_prompt_len()
         self._is_shutdown = False
         atexit.register(self.shutdown)
 
-    @property
-    def max_prompt_len(self) -> int | None:
-        """Longest prompt SGLang will read in full, or ``None`` if the context length is unknown.
+    def _resolve_max_prompt_len(self) -> int | None:
+        """Longest prompt this engine will read in full, or ``None`` if it can't be determined.
 
-        Prompts longer than this are silently truncated by the engine rather than rejected (see
-        :data:`_SGLANG_INPUT_RESERVE`), so callers that build their own prompt windows -- notably
-        :meth:`MEDS_EIC_AR.model.model.Model._rolling_generate` -- must clamp to this rather than
-        to the model's raw context length. ``None`` means "unknown, don't clamp", which restores
-        the pre-fix behavior rather than guessing a ceiling that might be wrong.
+        Asks the engine rather than re-deriving the arithmetic: ``scheduler_info`` carries the
+        ``max_req_input_len`` the scheduler actually enforces, so this tracks SGLang's own rule
+        across versions and accounts for the KV pool binding instead of the context length.
+        Subtracts one because SGLang's length check is ``>=``, not ``>``.
+
+        Falls back to :data:`_SGLANG_INPUT_RESERVE` against the model's context length when the
+        engine reports nothing (older SGLang, or a test double).
         """
+        info = getattr(self._engine, "scheduler_info", None)
+        reported = info.get("max_req_input_len") if isinstance(info, dict) else None
+        if isinstance(reported, int) and not isinstance(reported, bool) and reported > 0:
+            return reported - 1
         if self._context_len is None:
             return None
         return max(self._context_len - _SGLANG_INPUT_RESERVE, 0)
+
+    @property
+    def max_prompt_len(self) -> int | None:
+        """Longest prompt SGLang will read in full, or ``None`` if it could not be determined.
+
+        Prompts longer than this are silently truncated by the engine rather than rejected, so
+        callers that build their own prompt windows -- notably
+        :meth:`MEDS_EIC_AR.model.model.Model._rolling_generate` -- must size them against this
+        rather than the model's raw context length. ``None`` means "unknown, don't clamp".
+        """
+        return self._max_prompt_len
 
     def _check_prompt_len(self, prompt_len: int) -> None:
         """Refuse a prompt SGLang would silently truncate.
--- a/src/MEDS_EIC_AR/model/backends/sglang.py
+++ b/src/MEDS_EIC_AR/model/backends/sglang.py
@@ -109,6 +109,33 @@ _SGLANG_OUTPUT_KEYS: tuple[str, ...] = ("output_ids", "token_ids")
 _SGLANG_CONTEXT_RESERVE = 2
 
 
+#: Positions SGLang holds back from the *input* side of the context window.
+#:
+#: Distinct from :data:`_SGLANG_CONTEXT_RESERVE`, which bounds how many tokens SGLang will
+#: *emit*. This one bounds how many prompt tokens it will actually *read*, and exceeding it is
+#: far more damaging: SGLang does not refuse the request, it silently drops the tail of the
+#: prompt and generates from the truncated prefix, so the model is conditioned on something the
+#: caller never asked for. Derivation, from ``sglang==0.5.9``::
+#:
+#:     max_req_len       = min(context_len - 1, max_token_pool_size - 1)   # tp_worker.py
+#:     max_req_input_len = max_req_len - 5                                 # tp_worker.py
+#:     if len(input_ids) >= max_req_input_len: truncate or reject          # managers/utils.py
+#:
+#: The comparison is ``>=``, so the longest prompt SGLang reads in full is
+#: ``max_req_input_len - 1``, i.e. ``context_len - 7`` whenever the context length is the
+#: binding term. Measured against a 16-position model: a 9-token prompt is honored, a 10-token
+#: prompt is already truncated.
+#:
+#: Why this is not merely SGLang's business: ``allow_auto_truncate=True`` (set unconditionally
+#: below) is what turns the rejection into a silent truncation, and SGLang's own truncation
+#: warning is emitted through a logger its engine init has already set to ERROR, so nothing
+#: surfaces. Left unchecked, ``Model._rolling_generate``'s default ``rolling_context_size`` of
+#: ``max_seq_len - 1`` sits far above this ceiling and every rolling chunk loses its prompt
+#: tail — which is exactly how SGLang came to emit grammar-invalid trajectories where the HF
+#: backend emitted valid ones (issue #171).
+_SGLANG_INPUT_RESERVE = 7
+
+
 #: Smallest attention head dimension SGLang's default FlashInfer attention backend can dispatch.
 #: Below this it aborts the scheduler subprocess with ``FlashInfer Internal Error: Invalid
 #: configuration``, which reaches the parent as SIGQUIT / exit ``-9`` — indistinguishable at a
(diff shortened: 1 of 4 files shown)


mmcdermott:
Updated with both review points.

**1. The ceiling now comes from the engine.** `max_prompt_len` reads `scheduler_info["max_req_input_len"]` off the running engine instead of re-deriving `context_len - 7` from `config.json`. Confirmed populated at runtime:

```
{'status': 'ready', 'max_total_num_tokens': 34276968, 'max_req_input_len': 10}
```

(16-position model; matches the derived value.) Minus one, because SGLang's check is `>=`. This tracks SGLang's rule across versions rather than hard-coding its internal `-5`, and it covers what the derivation cannot express: when the **KV pool** binds instead of the context length, the true ceiling is smaller and depends on runtime memory. The arithmetic remains as a documented fallback for engines that report nothing.

**2. An explicitly-set `rolling_context_size` over the usable window is now an error.** Silently shrinking it means the run conditions on less history than it was told to, with nothing in the output saying so — and it makes bugs of exactly this class harder to reproduce on purpose, since asking for the oversized window that triggers them just hands you a smaller one. `None` still picks the largest usable window automatically, unchanged. The message names whichever bound binds, since a value the model's context could hold may still exceed the backend's:

```
rolling_context_size=15 exceeds what is usable here: SGLangBackend, which reads at most
9 prompt tokens and silently discards the rest. Lower it to at most 9, or leave it unset
to use the largest usable window automatically.
```

## Verification (DGX Spark, GB10 / SM 12.1)

| suite | result |
| --- | --- |
| `tests/grammar/` (both HF and SGLang backends) | **19 passed, 0 skipped** |
| `tests/`, `src/` doctests, `README.md` (excl. grammar) | 131 passed, 4 skipped |
| `tests/test_sglang_backend.py` | 39 passed (7 new) |

The SGLang grammar gate passes with default settings and its output is byte-identical to HF's.

## Re: running the gated test from pytest

It already behaves that way — no setup needed:

```bash
pytest tests/grammar/test_cli_sglang.py -rs
```

It runs when the environment supports it and skips otherwise, via `pytest.importorskip("sglang")` for the optional dependency and a `pytestmark = skipif(not torch.cuda.is_available())` for the GPU. `-rs` prints the reason, so a skip is visible rather than silent.

One papercut worth knowing: the grammar fixtures shell out to `MEICAR_process_data` / `MEICAR_pretrain`, so if the venv's `bin` is not on `PATH` the suite **errors** with `FileNotFoundError: 'MEICAR_process_data'` instead of skipping. That applies to every grammar test, not just this one. On this box SGLang additionally needs `.venv/bin` on `PATH` at runtime because it shells out to `ninja` for JIT kernels, so:

```bash
PATH=$PWD/.venv/bin:$PATH TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas \
  pytest tests/grammar/ -rs
```

Happy to convert that `FileNotFoundError` into a proper skip in a follow-up if you want the suite to degrade cleanly when the CLIs aren't on `PATH`.

src/MEDS_EIC_AR/model/backends/sglang.py
```diff
@@ -109,6 +109,25 @@
...
+#:     max_req_len       = min(context_len - 1, kv_pool_size - 1)   # managers/tp_worker.py
+#:     max_req_input_len = max_req_len - 5                          # managers/tp_worker.py
+#:     if len(input_ids) >= max_req_input_len: truncate             # managers/utils.py
+#:
+#: The comparison is ``>=``, so the longest prompt read in full is ``max_req_input_len - 1``,
+#: i.e. ``context_len - 7`` when the context length is the binding term. Prefer the engine's own
+#: ``scheduler_info["max_req_input_len"]`` (see :meth:`SGLangBackend._resolve_max_prompt_len`),
+#: which also covers the case where the KV pool binds instead of the context; this constant is
+#: only the fallback for engines that don't report it.
+_SGLANG_INPUT_RESERVE = 7
```
mmcdermott: Don't fall back to this. Given our plan now is to read this from the engine, that should always trump this, and if we don't see something in the engine but expect to, we should throw an error, as we don't know this is right in that case. We shouldn't have this parameter at all. We may also not need the other constant, but this certainly should go.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -897,12 +897,47 @@ def _rolling_generate(
...
+        # The prompt window is bounded by the model's context, and possibly further by the
+        # backend: some engines read fewer prompt tokens than the model can represent and
+        # silently discard the rest, which corrupts the conditioning rather than merely
+        # shortening it. Backends advertise that limit via an optional ``max_prompt_len``;
+        # ``None`` (the HF backend, which has no such limit) means the model context is the only
+        # bound.
+        ctx_cap = self.max_seq_len - 1
+        backend_cap = getattr(self._backend, "max_prompt_len", None)
+        if backend_cap is not None:
+            if backend_cap <= 0:
```
mmcdermott: This doesn't look right -- why are you comparing to 0 here? When would the backend_cap ever be zero? I feel like you should either be doing a true comparison here or an error should be raised upstream on the _backend when max_prompt_len is set.

src/MEDS_EIC_AR/model/model.py
```diff
@@ -897,12 +897,47 @@ def _rolling_generate(
...
+                    f"Backend {type(self._backend).__name__} reports max_prompt_len="
+                    f"{backend_cap}; it cannot accept any prompt for a model with "
+                    f"max_seq_len={self.max_seq_len}."
+                )
+            ctx_cap = min(ctx_cap, backend_cap)
+
         if rolling_context_size is None:
-            ctx_size = self.max_seq_len - 1
+            # Unspecified: take the largest window everything involved can honor.
+            ctx_size = ctx_cap
```
mmcdermott: Why do you need another ctx_size variable here? Why not just set rolling_context_size to the target?

src/MEDS_EIC_AR/model/backends/sglang.py
mmcdermott: Agreed, and done — `_SGLANG_INPUT_RESERVE` is gone.

Went further and removed `_SGLANG_CONTEXT_RESERVE` too, since you flagged it as a candidate. Both of SGLang's ceilings descend from one quantity, and the engine reports one of them:

```
max_req_len       = min(context_len - 1, kv_pool - 1)   # tp_worker.py
max_req_input_len = max_req_len - 5
```

So a single documented gap constant (`_SGLANG_MAX_REQ_LEN_GAP = 5`) recovers `max_req_len` from the reported value, and both ceilings follow. Verified against live engines: `max_req_input_len` is 506 on a 512-position model (→ 511 − 128 − 1 = 382, exactly the output ceiling I measured) and 10 on the 16-position grammar model.

There is no fallback any more: if the engine reports nothing usable, construction raises. Your reasoning is right — a guessed ceiling that is too high means the engine silently truncates prompts, and nothing downstream can tell. The test double now reports `scheduler_info` like a real engine, which also makes the mocks more faithful.

One thing the change buys beyond removing a constant: when the **KV pool** binds instead of the context length, the real ceiling is smaller and depends on runtime memory. The old config-derived arithmetic could not see that case at all.

src/MEDS_EIC_AR/model/model.py
mmcdermott: You are right on both counts — removed.

`backend_cap <= 0` was defensive noise in the wrong place. A backend that cannot accept any prompt is a fact the backend knows at construction, so that is where it is now checked: `SGLangBackend` raises if the engine reports a `max_req_input_len` leaving no room. `_rolling_generate` just takes the value.

That collapses the block to:

```python
backend_cap = getattr(self._backend, "max_prompt_len", None)
ctx_cap = self.max_seq_len - 1 if backend_cap is None else min(self.max_seq_len - 1, backend_cap)
```

src/MEDS_EIC_AR/model/model.py
mmcdermott: Fair — it does not earn its keep any more. It was left over from when this clamped; now that an explicit oversized value is an error, `rolling_context_size` and the target only differ in the `None` case, so I assign into it directly and drop `ctx_size` as a separate concept:

```python
if rolling_context_size is None:
    rolling_context_size = ctx_cap
elif rolling_context_size <= 0:
    ...
elif rolling_context_size > ctx_cap:
    ...
ctx_size = rolling_context_size
```

I kept the final `ctx_size = rolling_context_size` alias only because the loop below reads `ctx_size` in several places and renaming those is churn unrelated to this PR; happy to rename throughout if you would rather.

mmcdermott:
Review comments on all three PRs are addressed here, since #172 sits on top of the stack and the changes are entangled. #167 is merged and the stack is synced onto it (`README.md` auto-merged cleanly — the `mem_fraction_static` conflict I predicted did not materialise; #169's rewrite and #167's `uv sync` note both survived).

## Changes

**Both ceilings now come from the engine** (#172 and #168 threads). `scheduler_info["max_req_input_len"]` is the number the scheduler enforces; `max_req_len` follows from it via one documented gap constant. That deleted **both** `_SGLANG_INPUT_RESERVE` and `_SGLANG_CONTEXT_RESERVE`, and removed the config-derived context length entirely. Verified against live engines: 506 for a 512-position model (→ 511 − 128 − 1 = 382, the measured output ceiling) and 10 for the 16-position grammar model. No fallback — an engine reporting nothing usable raises rather than guessing.

**The attention-backend guard is hardened** (#169 thread). The head-dim floor now applies whenever FlashInfer will actually run, not only when it was left unnamed — naming it explicitly does not make it dispatch a narrow head dim. Unknown names are validated against SGLang's own `ATTENTION_BACKEND_CHOICES`. Non-FlashInfer backends below the floor are still allowed, which is what keeps `sglang_demo.yaml` working with the grammar fixture.

**Config reads go through `transformers.AutoConfig`** rather than hand-parsed JSON, so `head_dim` resolves even when derived from `hidden_size / num_attention_heads` — a case the old parser silently skipped.

**`_rolling_generate` simplified**: the `backend_cap <= 0` branch is gone (that condition is now rejected at construction, where it is known), and `rolling_context_size` is assigned into directly instead of introducing a second variable.

## Verification (DGX Spark, GB10 / SM 12.1)

| suite | result |
| --- | --- |
| `tests/grammar/` — both HF and SGLang | **19 passed, 0 skipped** |
| `tests/`, `src/` doctests, `README.md` | 132 passed, 4 skipped |
| `tests/test_sglang_backend.py` | 41 passed |

Re-ran the full grammar suite on GPU *after* the refactor, since the ceilings are computed differently now; SGLang output is still byte-identical to HF's. Pre-commit hooks run locally this time.

## Still open, not addressed here

The suppressed-warning problem I raised on #168: `sgl.Engine(...)` sets the root logger to ERROR, so the clamp warning in `_cap_max_new_tokens` (and the repo's own `Writing trajectory ...` INFO lines) never appear. That is unchanged — the warning survived the refactor but is still invisible in practice. It is a visibility bug rather than a correctness one, and the fix is independent of everything above, so I left it out to keep this diff reviewable. Happy to take it next, either here or as its own PR.

